PR #7, opened 2024-08-12
Dseat 12 create class model

lulli-110010100:
https://re-motion.atlassian.net/browse/DSEAT-12

### DIFF
--- a/RubeesEat.UnitTests/Model/BillTests.cs
+++ b/RubeesEat.UnitTests/Model/BillTests.cs
@@ -1,0 +1,53 @@
+using System.Collections.Immutable;
+using NUnit.Framework.Constraints;
+using RubeesEat.Model;
+
+namespace RubeesEat.UnitTests.Model;
+
+[TestFixture]
+public class BillTests
+{
+    [Test]
+    public void Constructor_AssignsPropertiesCorrectly()
+    {
+        Guid expectedGuid = Guid.NewGuid();
+        DateTime expectedTime = DateTime.Now;
+        ImmutableArray<EntryLine> entryLines =
+        [
+            new EntryLine(new Person(Guid.NewGuid(), "Max", "Ackermann"), 100),
+            new EntryLine(new Person(Guid.NewGuid(), "Minecraft", "Steve"), -50),
+            new EntryLine(new Person(Guid.NewGuid(), "Schmoes", "Relax"), -50)
+        ];
+        
+        Bill bill = new Bill(expectedGuid, expectedTime, "Going to Macdonal", entryLines);
+        
+        Assert.That(bill.Id, Is.EqualTo(expectedGuid));
+        Assert.That(bill.Date, Is.EqualTo(expectedTime));
+        Assert.That(bill.Description, Is.EqualTo("Going to Macdonal"));
+        Assert.That(bill.EntryLines.Length, Is.EqualTo(3));
+
+    }
+
+    [Test]
+    public void Constructor_WithSumNotZero_ThrowsException()
+    {
+        ImmutableArray<EntryLine> entryLines =
+        [
+            new EntryLine(new Person(Guid.NewGuid(), "Schmoes", "Relax"), 100),
+            new EntryLine(new Person(Guid.NewGuid(), "Klara", "Malernachzahler"), -50)
+        ];
+        
+        Assert.That(() => new Bill(Guid.NewGuid(), DateTime.Now, "Bolognese Flade in der Fladerei", entryLines), Throws.TypeOf<ArgumentException>().With.Message.EqualTo("Sum of entry lines should be 0."));
+    }
+
+    [Test]
+    public void Constructor_WithLessThanTwoEntryLines_ThrowsException()
+    {
+        ImmutableArray<EntryLine> entryLines =
+        [
+            new EntryLine(new Person(Guid.NewGuid(), "Doctor", "Doofenshmirtz"), 100)
+        ];
+        
+        Assert.That(() => new Bill(Guid.NewGuid(), DateTime.Now, "Alleine bei seiner Geburt", entryLines), Throws.TypeOf<ArgumentException>().With.Message.EqualTo("Entry lines should contain at least 2."));
+    }
+}
--- a/RubeesEat.UnitTests/Model/InMemoryPersonRepositoryTests.cs
+++ b/RubeesEat.UnitTests/Model/InMemoryPersonRepositoryTests.cs
@@ -1,0 +1,65 @@
+using System.Collections.Immutable;
+using RubeesEat.Model;
+
+namespace RubeesEat.UnitTests.Model;
+
+[TestFixture]
+public class InMemoryPersonRepositoryTests
+{
+    [Test]
+    public void GetAll_EntriesInDictionary_ReturnsEntries()
+    {
+        var storage = new Dictionary<Guid, Person>();
+        storage.Add(TestDomain.Jack.Id, TestDomain.Jack);
+        storage.Add(TestDomain.Erwin.Id, TestDomain.Erwin);
+        storage.Add(TestDomain.Sasha.Id, TestDomain.Sasha);
+        var personRepository = new InMemoryPersonRepository(TestDomain.Jack.Id, storage);
+        
+        var allEntries = personRepository.GetAll();
+        
+        Assert.That(allEntries.Count, Is.EqualTo(3));
+        Assert.That(allEntries[0].Id, Is.EqualTo(TestDomain.Jack.Id));
+        Assert.That(allEntries[1].Id, Is.EqualTo(TestDomain.Erwin.Id));
+        Assert.That(allEntries[2].Id, Is.EqualTo(TestDomain.Sasha.Id));
+    }
+
+    [Test]
+    public void Add_NoEntriesInDictionary_EntriesSuccessfullyAdded()
+    {
+        var storage = new Dictionary<Guid, Person>();
+        var personRepository = new InMemoryPersonRepository(TestDomain.Jack.Id, storage);
+        
+        personRepository.Add(TestDomain.Jack);
+        personRepository.Add(TestDomain.Erwin);
+        personRepository.Add(TestDomain.Sasha);
+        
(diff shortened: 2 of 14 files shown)


RubeesEat.UnitTests/Model/BillTests.cs
```diff
@@ -0,0 +1,41 @@
...
+    [Test]
+    public void Bill_CorrectParameters_DoesNotThrowException()
+    {
+        ImmutableArray<EntryLine> entryLines =
+        [
+            new EntryLine(new Person(Guid.NewGuid(), "test", "test"), 100),
+            new EntryLine(new Person(Guid.NewGuid(), "testb", "testb"), -50),
+            new EntryLine(new Person(Guid.NewGuid(), "testc", "testc"), -50)
+        ];
+        Assert.DoesNotThrow(() => new Bill(Guid.NewGuid(), DateTime.Now, "Test", entryLines));
```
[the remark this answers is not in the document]
FelixSchoberRubi: Why should we only use Assert.That? When you say "this is the new method", do you mean that Assert.DoesNotThrow is deprecated?

RubeesEat.UnitTests/Model/PersonTest.cs
```diff
@@ -0,0 +1,27 @@
+using RubeesEat.Model;
+
+namespace RubeesEat.UnitTests.Model;
+
+[TestFixture]
+public class PersonTest
+{
+    [Test]
+    public void Person_ValidParamenters_DoesNotThrowException()
```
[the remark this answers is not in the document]
IremA99: Why do we have to do this in PersonTests but not in BillTests?

RubeesEat.UnitTests/Model/PersonTest.cs
```diff
@@ -0,0 +1,27 @@
...
+    [Test]
+    [TestCase("Test2")]
+    [TestCase("3Test")]
+    [TestCase("?Test")]
+    [TestCase("1Test1")]
+    [TestCase("Te st")]
+    [TestCase(".")]
+    [TestCase("")]
+    [TestCase(" ")]
+    public void Person_NameInvalid_ThrowsException(string name)
```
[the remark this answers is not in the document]
IremA99: In BillTests you mentioned to write "Bill_ConstructorWithInvalidName_Throws.". Should we use the same convention here as well?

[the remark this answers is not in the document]
FelixSchoberRubi: I'm guessing we just have to be consistent. I will for now change it to

Bill_Constructor_WithSumNotZero_ThrowsException
Person_WithInvalidName_ThrowsException

We can change it if there is disagreement about this, but otherwise this should be consistent now.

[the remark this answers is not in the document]
FelixSchoberRubi: We agreed to
Bill_ConstructorWithSumNotZero_ThrowsException
Person_ConstructorWithInvalidName_ThrowsException

let us know if we need to change it

IremA99 (approved):

RubeesEat.UnitTests/Model/TestDomain.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace RubeesEat.UnitTests.Model;
+
+public static class TestDomain
+{
+    public static readonly List<Person> TestPersons =
+    [
+        Person.Create("DefaultUserFirstName", "DefaultUserLastName"), 
+        Person.Create("SecondUserFirstName", "SecondUserLastName"),
+        Person.Create("ThirdUserFirstName", "ThirdUserLastName"),
```
a-ctor, line 12: For asserts it might be better to have direct access to the persons. I would also give them actual personas to make the tests more readable. e.g. `TestDomain.Alice`, `TestDomain.Bob` etc. (instead of `TestDomain.TestPersons[0]`). The list of persons would then use these properties to fill the list. e.g. `TestDomain.AllPeople`, which contains all people instances.

RubeesEat.UnitTests/Model/TestDomain.cs
```diff
@@ -0,0 +1,34 @@
...
+    ];
+
+    public static readonly List<Bill> TestBill = [];
```
a-ctor, line 15: Name `TestBill` is a bit strange for a list. I would suggest `AllBills`. Please also add properties for the individual bill objects to allow their use in the test without indexer. You can also give them names like `BillFladerei` or something to differentiate them better in the tests.

RubeesEat.UnitTests/Model/PersonTest.cs
```diff
@@ -0,0 +1,34 @@
+using RubeesEat.Model;
+
+namespace RubeesEat.UnitTests.Model;
+
+[TestFixture]
+public class PersonTest
+{
+    [Test]
+    public void Person_Constructor_AssignsPropertiesCorrectly()
```
a-ctor, line 9: As discussed, please remove the `Person_` prefix that was a mistake by me from the older comment. We just want to use `Constructor` as name for the first part of the test. So we had `Name_Setup_Expectation` (e.g. `Constructor_WithInvalidParameter_ThrowsException`). In some cases we can remove `Setup` and `Expectation` if there is a clear cut standard method execution. In the case of the constructor, this standard case is that the constructor executed without exceptions we can use just `Constructor` as name. If we had multiple constructors we could use `Setup` to differentiate (e.g. `Constructor_WithoutParameters` and `Constructor_WithName`).

Please go through all the tests and fix the naming accordingly.

RubeesEat.UnitTests/Model/PersonTest.cs
```diff
@@ -0,0 +1,34 @@
...
+    {
+        Guid expectedGuid = Guid.NewGuid();
+        string expectedFirstName = "test";
+        string expectedLastName = "test";
+        var person = new Person(expectedGuid, expectedFirstName, expectedLastName);
+        Assert.That(person.Id, Is.EqualTo(expectedGuid));
+        Assert.That(person.FirstName, Is.EqualTo(expectedFirstName));
+        Assert.That(person.LastName, Is.EqualTo(expectedLastName));
```
a-ctor, line 17: Having a bunch of lines clumped together like that makes the code harder to read. We can use blank lines to separate parts of the method that do separate things. In the case of a test the most obvious case are the parts `arrange - act - assert`. Blank lines between these parts make the test a bit more readable.

In tests, we can prefer readability over repetition. As such, I would suggest inlining the `expectedFirstName` and `expectedLastName`. There is also a problem with the values you chose. As first and last name are the same value, we don't test that first and last name are assigned to the correct field. So we might have `_firstName = lastName;` in the constructor and the test still passes. Also, human readable names that fit the context are better than random test string. So using `Max Muster`, for example, is better than `test test2` :p

With these changes we arrive at something like this:

```suggestion
        Guid expectedGuid = Guid.NewGuid();
        
        var person = new Person(expectedGuid, "Max", "Muster");
        
        Assert.That(person.Id, Is.EqualTo(expectedGuid));
        Assert.That(person.FirstName, Is.EqualTo("Max"));
        Assert.That(person.LastName, Is.EqualTo("Muster"));
```

Please go through the other tests and apply the same things there to make them more readable and simplify them.

RubeesEat.UnitTests/Model/PersonTest.cs
```diff
@@ -0,0 +1,34 @@
...
+    [TestCase("3Test")]
+    [TestCase("?Test")]
+    [TestCase("1Test1")]
+    [TestCase("Te st")]
+    [TestCase(".")]
+    [TestCase("")]
+    [TestCase(" ")]
+    public void Person_ConstructorWithInvalidName_ThrowsException(string name)
+    {
+        Assert.That(() => new Person(Guid.NewGuid(), name, name), Throws.TypeOf<ArgumentException>().With.Message.EqualTo("Name must only consist of letters."));
```
a-ctor, line 32: When lines become too long it is better for readability to split them into multiple lines. I would suggest splitting up all parameters on new lines and then splitting up the `.With.Message...` thing on its own line like this:

```suggestion
        Assert.That(
            () => new Person(Guid.NewGuid(), name, name), 
            Throws.TypeOf<ArgumentException>()
                .With.Message.EqualTo("Name must only consist of letters."));
```

RubeesEat.UnitTests/Model/InMemoryPersonRepositoryTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    [Test]
+    public void GetCurrentUser_IsInDictionary_ReturnsCurrentUser()
+    {
+        Person defaultUser = new Person(Guid.NewGuid(), "DefaultFirstName", "DefaultLastName");
+
+        var personRepository = new InMemoryPersonRepository(defaultUser.Id);
+        personRepository.Add(defaultUser);
+
+        Assert.That(personRepository.GetCurrentUser(),
+            Is.EqualTo(defaultUser)); // in this case reference equals is fine because its the same object
```
a-ctor, line 46: I don't think that the comments add any value and should be removed.

RubeesEat.UnitTests/Model/InMemoryPersonRepositoryTests.cs
```diff
@@ -0,0 +1,61 @@
...
+    public void GetCurrentUser_IsNotInDictionary_ThrowsException()
+    {
+        Person defaultUser = new Person(Guid.NewGuid(), "DefaultFirstName", "DefaultLastName");
+
+        var personRepository = new InMemoryPersonRepository(defaultUser.Id);
+
+        Assert.That(() => personRepository.GetCurrentUser(),
+            Throws.TypeOf<ArgumentException>().With.Message
+                  .EqualTo(
+                      "default User is not in the dictionary")); // in this case reference equals is fine because its the same object
```
a-ctor, line 59: I don't think that the comments add any value and should be removed.

When splitting a method call into multiple lines I would suggest moving each paramter on their own line:

```suggestion
        Assert.That(
            () => personRepository.GetCurrentUser(),
            Throws.TypeOf<ArgumentException>().With.Message
                  .EqualTo(
                      "default User is not in the dictionary")); // in this case reference equals is fine because its the same object
```

a-ctor (changes requested):

a-ctor (approved):

PR #8, opened 2024-08-14
Feature/dseat 15 get latest balance changes

lulli-110010100:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-15

### DIFF
--- a/RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
+++ b/RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
@@ -1,0 +1,34 @@
+using RubeesEat.IntegrationTests.PageObjects;
+
+namespace RubeesEat.IntegrationTests.SeleniumTests;
+
+[TestFixture]
+public class RubeesEatSeleniumTests : SeleniumIntegrationTestBase
+{
+    public RubeesEatSeleniumTests() : base("UserHomepage")
+    {
+    }
+    
+    [Test]
+    public void GetDefaultBalanceChanges()
+    {
+        var page = Start<UserHomepagePageObject>();
+        var balanceChanges = page.GetBalanceChanges();
+
+        Assert.That(balanceChanges[0].Amount, Is.EqualTo("1000€"));
+        Assert.That(balanceChanges[0].Date, Is.EqualTo("Tuesday, August 19, 2025"));
+        Assert.That(balanceChanges[0].Description, Is.EqualTo("Default user paid lunch 2025"));
+        
+        Assert.That(balanceChanges[1].Amount, Is.EqualTo("-800€"));
+        Assert.That(balanceChanges[1].Date, Is.EqualTo("Sunday, April 28, 2024"));
+        Assert.That(balanceChanges[1].Description, Is.EqualTo("Lunch with default user in 2024"));
+        
+        Assert.That(balanceChanges[2].Amount, Is.EqualTo("-800€"));
+        Assert.That(balanceChanges[2].Date, Is.EqualTo("Friday, April 28, 2023"));
+        Assert.That(balanceChanges[2].Description, Is.EqualTo("Lunch with default user in 2023"));
+        
+        Assert.That(balanceChanges[3].Amount, Is.EqualTo("-800€"));
+        Assert.That(balanceChanges[3].Date, Is.EqualTo("Thursday, April 28, 2022"));
+        Assert.That(balanceChanges[3].Description, Is.EqualTo("Lunch with default user 2022"));
+    }
+}
--- a/RubeesEat.UnitTests/Model/InMemoryBillRepositoryTests.cs
+++ b/RubeesEat.UnitTests/Model/InMemoryBillRepositoryTests.cs
@@ -13,25 +13,201 @@ public class InMemoryBillRepositoryTests
         storage.Add(TestDomain.BillCafeLeBlanc.Id, TestDomain.BillCafeLeBlanc);
         storage.Add(TestDomain.BillMaidCafe.Id, TestDomain.BillMaidCafe);
         var billRepository = new InMemoryBillRepository(storage);
-        
+
         var allEntries = billRepository.GetAll();
-        
+
         Assert.That(allEntries.Count, Is.EqualTo(2));
         Assert.That(allEntries[0].Id, Is.EqualTo(TestDomain.BillCafeLeBlanc.Id));
         Assert.That(allEntries[1].Id, Is.EqualTo(TestDomain.BillMaidCafe.Id));
     }
-    
-    [Test] 
+
+    [Test]
     public void Add_NoEntriesInDictionary_EntriesSuccessfullyAdded()
     {
         var storage = new Dictionary<Guid, Bill>();
         var billRepository = new InMemoryBillRepository(storage);
-        
+
         billRepository.Add(TestDomain.BillCafeLeBlanc);
         billRepository.Add(TestDomain.BillMaidCafe);
-        
+
         Assert.That(billRepository.GetById(TestDomain.BillCafeLeBlanc.Id), Is.EqualTo(TestDomain.BillCafeLeBlanc));
         Assert.That(billRepository.GetById(TestDomain.BillMaidCafe.Id), Is.EqualTo(TestDomain.BillMaidCafe));
+    }
 
+    [Test]
+    public void GetAllFromUser_EntriesExist_ReturnsUserEntries()
+    {
+        Person defaultUser = Person.Create("DefaultFirstName", "DefaultLastName");
+        var billRepository = new InMemoryBillRepository();
+
+        var personRepository = new InMemoryPersonRepository(defaultUser.Id);
+        personRepository.Add(defaultUser);
+        personRepository.Add(new Person(Guid.NewGuid(), "Item", "Arslan"));
+        personRepository.Add(new Person(Guid.NewGuid(), "Patrick", "Widener"));
+        var persons = personRepository.GetAll().ToImmutableList();
+        var firstBill = new Bill(Guid.NewGuid(), new DateTime(2024, 9, 9), "Mittagessen auf Patrick sein Nacken",
+        [
+            new EntryLine(persons[0], 500),
+            new EntryLine(persons[1], -500),
+        ]);
+        billRepository.Add(firstBill);
+        billRepository.Add(new Bill(Guid.NewGuid(), new DateTime(2024, 9, 9), "Mittagessen auf Patrick sein Nacken",
+        [
+            new EntryLine(persons[1], 1_000_000),
+            new EntryLine(persons[2], -500_000),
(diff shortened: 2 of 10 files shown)


Ottorium (dismissed):

RubeesEat.UnitTests/Model/BillTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [
+            new EntryLine(new Person(Guid.NewGuid(), "test", "test"), 100),
+            new EntryLine(new Person(Guid.NewGuid(), "testb", "testb"), -50),
+            new EntryLine(new Person(Guid.NewGuid(), "testc", "testc"), -50)
+        ];
+        Assert.DoesNotThrow(() => new Bill(Guid.NewGuid(), DateTime.Now, "Test", entryLines));
+    }
+
+    [Test]
+    public void Bill_SumZero_ThrowsException()
```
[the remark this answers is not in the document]
lulli-110010100: fixen wir beim merge, in create class model ist es dann richtig

RubeesEat/Pages/UserHomepage.cshtml
```diff
@@ -0,0 +1,21 @@
...
+<ul id="balanceChanges" style="list-style: none">
+    @foreach (var item in billRepository.GetRecentBalanceChanges(personRepository.GetCurrentUser().Id, 10))
+    {
+        @*@item.Amount < 0*@
+        <li>
+
+            <div style="display: inline; color: @(item.Amount < 0 ? "red" : "green");">@item.Amount€</div>
+
+            @item.Date.DayOfWeek
+            @item.Date.ToString("dd/MM/yyyy")
```
a-ctor, line 17: I would suggest using a format specifier that already includes the the day of the week like `"D"` (see https://learn.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings). This way, we only have one statement to format the date and it uses the current culture to format the date.

RubeesEat/Startup.cs
```diff
@@ -26,22 +26,49 @@ public void ConfigureServices(IServiceCollection services)
...
+        var billRepository = new InMemoryBillRepository();
         personRepository.Add(defaultUser);
         personRepository.Add(new Person(Guid.NewGuid(), "Item", "Arslan"));
         personRepository.Add(new Person(Guid.NewGuid(), "Patrick", "Widener"));
         personRepository.Add(new Person(Guid.NewGuid(), "Lilli", "Grubber"));
         personRepository.Add(new Person(Guid.NewGuid(), "Mich", "Ludwig"));
 
         var persons = personRepository.GetAll().ToImmutableList();
-        var billRepository = new InMemoryBillRepository();
         billRepository.Add(new Bill(Guid.NewGuid(), new DateTime(2024, 9, 9), "Mittagessen auf Patrick sein Nacken",
```
a-ctor, line 40: These seem to be merge artifacts that should be reverted.

RubeesEat/Startup.cs
```diff
@@ -61,11 +88,10 @@ public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
...
-            {
-                endpoints.MapRazorPages();
-                endpoints.MapControllers();
-            });
+        app.UseEndpoints(endpoints =>
+        {
+            endpoints.MapRazorPages();
+            endpoints.MapControllers();
+        });
     }
```
a-ctor, line 96: revert please

RubeesEat/Model/IBillRepository.cs
```diff
@@ -4,4 +4,6 @@ public interface IBillRepository
 {
     IReadOnlyList<Bill> GetAll();
     void Add(Bill bill);
+    IReadOnlyList<Bill> GetAllFromUser(Guid userGuid);
```
a-ctor, line 7: I think `For` instead of `From` would improve the name

RubeesEat/Model/Bill.cs
```diff
@@ -2,7 +2,7 @@
 
 namespace RubeesEat.Model;
 
-public class Bill
+public class Bill : IComparable<Bill>
```
a-ctor, line 5: Please revert the comparability here. See the other comment on how we can sort without it.

RubeesEat/Controllers/BillController.cs
```diff
@@ -0,0 +1,6 @@
+namespace RubeesEat.Controllers;
+
+public class BillController
+{
+    
+}
```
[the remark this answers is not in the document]
a-ctor: So why is it still here? :D

RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
```diff
@@ -0,0 +1,21 @@
...
+    }
+    
+    [Test]
+    public void GetDefaultBalanceChanges()
+    {
+        var page = Start<UserHomepagePageObject>();
+
+        Assert.That(
+            page.GetBalanceChangeTexts(),
+            Is.EqualTo(new[] { "1000€ Tuesday 19.08.2025 Default user paid lunch 2025", "-800€ Sunday 28.04.2024 Lunch with default user in 2024", "-800€ Friday 28.04.2023 Lunch with default user in 2023", "-800€ Thursday 28.04.2022 Lunch with default user 2022" }));
```
a-ctor, line 19: Ideally we don't just assert the text but also the structure. That requires us to render the elements (amount, description, date) in separate HTML elements that we can select. We can then retrieve these fields and construct a `BalanceChange` object in the page object. This allows us to do this assert on objects and no longer assert text here.

a-ctor (approved):

RubeesEat/Model/IBillRepository.cs
```diff
@@ -4,4 +4,6 @@ public interface IBillRepository
 {
     IReadOnlyList<Bill> GetAll();
     void Add(Bill bill);
+    IReadOnlyList<Bill> GetAllForUser(Guid userGuid);
+    IReadOnlyList<BalanceChange> GetRecentBalanceChanges(Guid userGuid, int amount);
```
a-ctor, line 8: Just noticed, we should be using the domain objects not the IDs here. So please change the two methods to take a `Person` instead.

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -29,4 +29,32 @@ public void Add(Bill bill)
         ArgumentNullException.ThrowIfNull(bill);
         _bills.Add(bill.Id, bill);
     }
+
+    public IReadOnlyList<Bill> GetAllForUser(Guid userGuid)
+    {
+        return (from bill in _bills
+            where Enumerable.Any(bill.Value.EntryLines, line => line.Person.Id == userGuid)
+            select bill.Value).ToList();
```
a-ctor, line 37: There two kind of writing LINQ queries. One is this query syntax you've used here. This is similar to SQL and allows stuff like filtering, selection, grouping, joining, etc.

The other way is using the method syntax. See this example on the differences in style: https://learn.microsoft.com/en-us/dotnet/csharp/linq/get-started/write-linq-queries#standard-query-operator-extension-methods

We don't really use the query syntax but instead use the method syntax. So your example would be:

```c#
return _bills.Values
    .Where(e => e.EntryLines.Any(f => f.Person == person)
    .ToList();
```

In this context I want you to take a look at two concepts that are relevant here:

 - The interface `IEnumerable<T>` and how to use it
 - Extension methods (https://learn.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/extension-methods) and how they relate to LINQ

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -29,4 +29,32 @@ public void Add(Bill bill)
...
+    }
+
+    public IReadOnlyList<BalanceChange> GetRecentBalanceChanges(Guid userGuid, int amount)
+    {
+        var billsWithUser = GetAllForUser(userGuid).ToList();
+        billsWithUser = billsWithUser.OrderByDescending(bill => bill.Date).ToList();
```
a-ctor, line 43: Once you've look at LINQ and `IEnumerable<T>` a bit can you tell me what seems wrong to me here?

RubeesEat/Model/Bill.cs
```diff
@@ -30,4 +30,5 @@ public Bill(Guid id, DateTime date, string description, ImmutableArray<EntryLine
         Description = description;
         EntryLines = entryLines;
     }
+
```
a-ctor, line 33: 👀

RubeesEat.UnitTests/Model/InMemoryBillRepositoryTests.cs
```diff
@@ -13,25 +13,206 @@ public void GetAll_EntriesInDictionary_ReturnsEntries()
...
+            new EntryLine(persons[1], -500_000)
+        ]));
+
+        var balanceChangesFromUser = billRepository.GetRecentBalanceChanges(defaultUserGuid, 1);
+
+        Assert.That(balanceChangesFromUser.Count, Is.EqualTo(0));
+    }
+
+    [Test]
+    public void GetAllFromUser_EntriesExist_OrderedByDate()
```
a-ctor, line 182: I think this test method is not named correctly as it tests `GetRecentBalanceChanges` and not `GetAllFromUser`

RubeesEat/Pages/UserHomepage.cshtml
```diff
@@ -0,0 +1,21 @@
...
+
+<ul id="balanceChanges" style="list-style: none">
+    @foreach (var item in billRepository.GetRecentBalanceChanges(personRepository.GetCurrentUser().Id, 10))
+    {
+        @*@item.Amount < 0*@
+        <li>
+
+            <div class="amount" style="display: inline; color: @(item.Amount < 0 ? "red" : "green");">@item.Amount€</div>
+
+            <div class="date" style="display: inline;">@item.Date.ToString("D", CultureInfo.CreateSpecificCulture("en-US"))</div>
```
a-ctor, line 17: Specifying a culture here might not be what we want. In the future we could allow users to change their preferred language. In such scenarios, we want the date to be formatted in the users culture. When you don't specify a culture, it will use the current culture of the PC to format the string.

RubeesEat/Model/InMemoryBillRepository.cs
FelixSchoberRubi: Is the takeaway, that we didn't need to make it into a List because IEnumerable already has OrderByDescending?

RubeesEat/Pages/UserHomepage.cshtml
FelixSchoberRubi: We have an issue during testing. Some of us get English Culture automatically, some German. So when we run the tests, we get different results.

a-ctor: Okay then we leave it as is an will handle this once we add localization :)

PR #10, opened 2024-08-22
feature/DSEAT-13 split a bill

FelixSchoberRubi:
Branch name was wrong, had to rename. This required a new pull request.

### DIFF
--- a/RubeesEat/Controllers/BillController.cs
+++ b/RubeesEat/Controllers/BillController.cs
@@ -1,0 +1,96 @@
+using System.Globalization;
+using Microsoft.AspNetCore.Http.HttpResults;
+using Microsoft.AspNetCore.Mvc;
+using RubeesEat.Model;
+
+namespace RubeesEat.Controllers;
+
+[ApiController]
+[Route("/api/bills")]
+public class BillController(IBillRepository billRepository, IPersonRepository personRepository)
+    : ControllerBase
+{
+    [Route("")]
+    [HttpGet]
+    public IEnumerable<Bill> GetAll()
+    {
+        return billRepository.GetAll();
+    }
+
+    [Route("")]
+    [HttpPost]
+    public Results<BadRequest<string>, RedirectHttpResult> Create()
+    {
+        var form = HttpContext.Request.Form;
+        string? description = form["billDescription"];
+        CultureInfo culture = new CultureInfo("en-EN");
+        if (description == null)
+        {
+            return TypedResults.BadRequest("Description was empty.");
+        }
+
+        List<EntryLine> entryLines = [];
+        if (!decimal.TryParse(form["billAmount"], culture, out decimal totalAmount))
+        {
+            return TypedResults.BadRequest("Total amount must be a number.");
+        }
+        entryLines.Add(new EntryLine(personRepository.GetCurrentUser(), totalAmount));
+
+        for (var i = 0;; i++)
+        {
+            string? id = form["id" + i];
+
+            if (id == null)
+            {
+                break;
+            }
+
+            Person? person;
+            if (!Guid.TryParse((id), out var personGuid))
+            {
+                return TypedResults.BadRequest("Person Guid not valid");
+            }
+
+            if (!personRepository.TryGetById(personGuid, out person))
+            {
+                return TypedResults.BadRequest("Person not found");
+            }
+
+            var stringValues = form["amount" + i];
+
+            if (!decimal.TryParse(stringValues, culture,out var amount))
+            {
+                return TypedResults.BadRequest("Amount of a person must be a number.");
+            }
+            
+            entryLines.Add(new EntryLine(person, -amount));
+        }
+        
+        if (entryLines[0].Amount <= 0)
+        {
+            return TypedResults.BadRequest("First entry should always be positive");
+        }
+
+        for(int i = 1; i < entryLines.Count; i++)
+        {
+            if (entryLines[i].Amount >= 0)
+            {
+                return TypedResults.BadRequest("Entries after the first one should always be negative");
+            }
+        }
+        
+        int linesAfterFirst = entryLines.Skip(1).DistinctBy(p => p.Person).Count();
+        int distinctLinesAfterFirst = entryLines.Skip(1).Count();
+        
+        if (linesAfterFirst != distinctLinesAfterFirst)
+        {
+            return TypedResults.BadRequest("Persons should not appear multiple times (other than the buyer)");
+        }
+
+        var bill = Bill.Create(description, entryLines.ToArray());
+
+        billRepository.Add(bill);
+
+        return TypedResults.Redirect("/SplitBill");
+    }
+}
--- a/RubeesEat/Controllers/PersonController.cs
+++ b/RubeesEat/Controllers/PersonController.cs
@@ -1,0 +1,16 @@
+using Microsoft.AspNetCore.Mvc;
+using RubeesEat.Model;
+
+namespace RubeesEat.Controllers;
+
+[ApiController]
+[Route("/api/persons")]
+public class PersonController(IPersonRepository personRepository) : ControllerBase
+{
+    [Route("")]
+    [HttpGet]
+    public IEnumerable<Person> GetAll()
+    {
+        return personRepository.GetAll();
+    }
+}
--- a/RubeesEat/Model/IPersonRepository.cs
+++ b/RubeesEat/Model/IPersonRepository.cs
@@ -1,3 +1,5 @@
+using System.Diagnostics.CodeAnalysis;
+
 namespace RubeesEat.Model;
 
 public interface IPersonRepository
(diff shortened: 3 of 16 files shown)


a-ctor (changes requested):

RubeesEat/Controllers/PersonController.cs
```diff
@@ -0,0 +1,16 @@
+using Microsoft.AspNetCore.Mvc;
+using RubeesEat.Model;
+
+namespace RubeesEat.Controllers;
+
+[ApiController]
+[Route("/api/persons")]
+public class PersonController(IPersonRepository personRepository) : ControllerBase
```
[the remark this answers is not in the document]
IremA99: We thought that we might need it in the future. Should we delete it?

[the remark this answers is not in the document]
a-ctor: If we don't need it now we can create it in the future once we need it. [YAGNI](https://en.wikipedia.org/wiki/You_aren%27t_gonna_need_it)

RubeesEat/Model/IPersonRepository.cs
```diff
@@ -5,4 +5,5 @@ public interface IPersonRepository
     IReadOnlyList<Person> GetAll();
     void Add(Person person);
     Person GetCurrentUser();
+    Person GetById(Guid guid);
```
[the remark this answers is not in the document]
FelixSchoberRubi: Should we also do that for GetCurrentUser (line 7)
In case the Current User gets deleted?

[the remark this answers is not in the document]
a-ctor: No that is not necessary. We only allow access to the application with authentication. As such, `GetCurrentUser` should always return a `Person` object. If the user is deleted while somebody is still logged in as one, we should throw an exception.

RubeesEat/Controllers/PersonController.cs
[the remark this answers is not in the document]
FelixSchoberRubi: It turns out we needed it for a selenium test. But only for that

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -0,0 +1,87 @@
...
+            selectElement = document.querySelector('#billPeople');
+            let selectedPerson = selectElement.value;
+            let selectedId = selectElement.options[selectElement.selectedIndex].getAttribute("data-id");
+            
+            let personSpan = document.createElement("span");
+            personSpan.textContent = selectedPerson;
+            personSpan.id = selectedId;
+            document.querySelector("#addedPeople").appendChild(personSpan);
+            
+            let id = document.createElement("input");
```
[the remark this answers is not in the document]
FelixSchoberRubi: is personId ok?

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -0,0 +1,87 @@
...
+            
+            var dropdown = document.getElementById("billPeople");
+            dropdown.remove(dropdown.selectedIndex)
+        }
+        
+        function validate() {
+            let amounts = document.querySelectorAll("[name^='amount']");
+            let description = document.getElementById("billDescription");
+            let sum = 0;
+            amounts.forEach((amount) => sum += parseFloat(amount.value));
```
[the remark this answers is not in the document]
IremA99: Why is it better to do it this way?

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -0,0 +1,87 @@
...
+    <script type="text/javascript">
+        let counter = 0;
+        function getOption() {
+            selectElement = document.querySelector('#billPeople');
+            let selectedPerson = selectElement.value;
+            let selectedId = selectElement.options[selectElement.selectedIndex].getAttribute("data-id");
+            
+            let personSpan = document.createElement("span");
+            personSpan.textContent = selectedPerson;
+            personSpan.id = selectedId;
```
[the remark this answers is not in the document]
FelixSchoberRubi: we needed it in line 49 and line 50 to append the input field "amount". We changed the id now to be person0, person1 etc. to be consistent with the rest of the function

RubeesEat/Controllers/BillController.cs
```diff
@@ -0,0 +1,97 @@
...
+        }
+
+        bool hasDuplicates = entryLines
+                             .Skip(1)
+                             .GroupBy(p => p.Person)
+                             .Any(g => g.Count() > 1);
+        if (hasDuplicates)
+        {
+            return TypedResults.BadRequest("Persons should not appear multiple times (other than the buyer)");
+        }
```
Ottorium, line 89: use .Distinct()

RubeesEat/Model/InMemoryPersonRepository.cs
```diff
@@ -35,9 +37,18 @@ public Person GetCurrentUser()
...
+            person = _persons[guid];
+            return true;
+        }
+        catch (KeyNotFoundException e)
+        {
+            person = null;
+            return false;
+        }
     }
 }
```
Ottorium, line 54: To be consistent, just do getByID returning Person?

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -0,0 +1,114 @@
...
+            
+            if (totalAmount !== sum) 
+            {
+                document.getElementById("placeForErrorMessage").textContent = "The sum of individual amounts doesn't match the total bill amount.";
+                return;
+            }
+
+            document.getElementById("splitBillForm").submit();
+        }
+    </script>
```
Ottorium, line 113: move to seperate file

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -0,0 +1,114 @@
...
+            });
+            if (amountsInvalid) {
+                document.getElementById("placeForErrorMessage").textContent = "Amounts are not valid.";
+                return;
+            }
+            
+            if (amountsNotPositive) {
+                document.getElementById("placeForErrorMessage").textContent = "Amounts should be positive.";
+                return;
+            }
```
Ottorium, line 79:
```suggestion
            amounts.forEach((amount) =>
            {
               if (!amount.value || amount.value <= 0) {
                    document.getElementById("placeForErrorMessage").textContent = "Amounts are not valid.";
                return;
                }
            });
```

Ottorium: adjust message to descript what am amount should look like and print the wrong amount

Ottorium (changes requested):

RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
```diff
@@ -27,38 +27,76 @@ public void CorrectInput()
...
+    
+    [Test]
+    public void NegativeAmount()
+    {
+        var page = Start<SplitBillPageObject>();
+        
+        page.SetNewDesciptionText("Test");
+        page.SetNewTotalPriceText("-10");
+        var person = page.ClickAddPerson();
+        person.SetAmountForPerson("-10");
```
Ottorium, line 87: amounts should add up to 0

Ottorium (changes requested):

RubeesEat/Controllers/BillController.cs
FelixSchoberRubi: according to this https://learn.microsoft.com/de-de/dotnet/api/system.linq.enumerable.distinct?view=net-8.0
.Distinct only returns a list of distinct elements. It does not tell me if the list only has distinct elements

FelixSchoberRubi: After discussion with Ottorium, we will compare the length of the list from Distinct with the original list

RubeesEat/Pages/SplitBill.cshtml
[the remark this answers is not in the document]
a-ctor: Yes but you already have the element as a variable in `personSpan`. So you don't need to give it an id to query the element later again.

RubeesEat/Pages/SplitBill.cshtml
IremA99: We cant do that because return does not end the function

IremA99: We cant always print the wrong amount because the html input element only sends numbers. e or --3 for example cant be depicted. Only negative numbers can be depicted.

RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
IremA99: -10 is the user input. The second input will be changed to 10 in EntryLine. So it already adds up to 0

RubeesEat/Pages/SplitBill.cshtml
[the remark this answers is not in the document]
FelixSchoberRubi: Oh, right! I removed it and used the variable

Ottorium (approved):

a-ctor (approved):

PR #11, opened 2024-08-22
Feature/dseat 14 get account balance

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-14

### DIFF
--- a/RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
+++ b/RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
@@ -3,12 +3,8 @@ using RubeesEat.IntegrationTests.PageObjects;
 namespace RubeesEat.IntegrationTests.SeleniumTests;
 
 [TestFixture]
-public class RubeesEatSeleniumTests : SeleniumIntegrationTestBase
+public class RubeesEatSeleniumTests() : SeleniumIntegrationTestBase("UserHomepage")
 {
-    public RubeesEatSeleniumTests() : base("UserHomepage")
-    {
-    }
-    
     [Test]
     public void GetDefaultBalanceChanges()
     {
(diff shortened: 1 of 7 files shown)


RubeesEat/Pages/UserHomepage.cshtml
```diff
@@ -0,0 +1,17 @@
...
+@inject IPersonRepository personRepository
+
+Hello, @personRepository.GetCurrentUser().FirstName!
+
+@{
+    var amount = billRepository.GetBalance(personRepository.GetCurrentUser().Id);
+    var isPositive = amount >= 0;
+    var displayColor = isPositive ?  "green" : "red";
+    var sign = isPositive ?  "+" : "-";
+    var amountAsString = sign + " " + amount.ToString("F2") + " €";
```
lulli-110010100, line 14: If you use the `var amountAsString = sign + " " + amount.ToString("C");` format, it looks better + it automatically adds the currency

RubeesEat.IntegrationTests/SeleniumTests/RubeesEatSeleniumTests.cs
```diff
@@ -0,0 +1,20 @@
...
+    }
+
+    [Test]
+    public void GetAccountBalance()
+    {
+        var page = Start<UserHomepagePageObject>();
+        Assert.That(
+            page.GetAccountBalance(),
+            Is.EqualTo("+ 1000000,50 \u20ac"));
+    }
```
lulli-110010100, line 19: `Is.EqualTo("+ 1000000,50 \u20ac"));` needs to be adapted if you use the currency format

lulli-110010100 (changes requested):

RubeesEat/Pages/UserHomepage.cshtml
FelixSchoberRubi: Currency should always be Euro. C changes to a different Currency in a different country, depending on settings.

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -29,4 +29,11 @@ public void Add(Bill bill)
         ArgumentNullException.ThrowIfNull(bill);
         _bills.Add(bill.Id, bill);
     }
+
+    public decimal GetBalance(Guid personId)
+    {
+        return _bills.Values.Sum(bill => bill.EntryLines
+                                             .Where(entryLine => entryLine.Person.Id == personId)
+                                             .Sum(entryLine => entryLine.Amount));
```
a-ctor, line 37: Alternative LINQ expression using `.SelectMany(...)` (`flatMap` in JS) as an example:

```c#
return _bills.Values
    .SelectMany(e => e.EntryLines)
    .Where(e => e.Person == person)
    .Sum(e => e.Amount);
```

`SelectMany` selects an `IEnumerable<T>` and "flattens" it out so that the elements for a single enumerable. So we change the enumeration from `Bill` to `EntryLine` because we don't care about the details of the `Bill`. Then we filter and sum just on an enumerable of `EntryLine`.

Example from JS, which might convey the idea a bit easier: `[[1, 3], 4, [5, 6]].flatMap(e => e) ` -> `[ 1, 3, 4, 5, 6 ]`

RubeesEat/Startup.cs
```diff
@@ -38,8 +38,8 @@ public void ConfigureServicesEAT(IServiceCollection services)
         var billRepository = new InMemoryBillRepository();
         billRepository.Add(new Bill(Guid.NewGuid(), new DateTime(2024, 9, 9), "Mittagessen auf Patrick sein Nacken",
         [
-            new EntryLine(persons[1], 1_000_000),
-            new EntryLine(persons[2], -500_000),
+            new EntryLine(persons[0], (decimal)1_000_000.50),
+            new EntryLine(persons[2], (decimal)-500_000.50),
```
a-ctor, line 42: In this case I would prefer adding a type suffix instead of doing a cast:

```suggestion
            new EntryLine(persons[0], 1_000_000.50m),
            new EntryLine(persons[2], 500_000.50m),

```

`m` stands for `decimal` (**m**oney)

a-ctor (approved):
Looks good, well done :)

a-ctor (approved):

PR #15, opened 2024-08-29
Feature/dseat 23 master layout

Ottorium:
https://re-motion.atlassian.net/browse/DSEAT-23


RubeesEat/Pages/UserHomepage.cshtml
```diff
@@ -3,33 +3,35 @@
...
 
-Hello, @personRepository.GetCurrentUser().FirstName!
+<h1 id="hello">Hello, @personRepository.GetCurrentUser().FirstName!</h1>
 
 @{
     var amount = billRepository.GetBalance(personRepository.GetCurrentUser());
     var isPositive = amount >= 0;
-    var displayColor = isPositive ? "green" : "red";
     var sign = isPositive ? "+" : "-";
     var amountAsString = sign + " " + amount.ToString("C", CultureInfo.CreateSpecificCulture("de-DE"));
```
IremA99, line 12: I think in the other branches we used the english culture. We should decide which one to use.

RubeesEat/wwwroot/css/UserHomePage.css
```diff
@@ -0,0 +1,118 @@
...
+
+@media (max-width: 480px) {
+    #hello {
+        font-size: 1.5em;
+    }
+
+    #balance {
+        font-size: 1.5em;
+    }
+}
```
FelixSchoberRubi, line 95: I would put the @media selectors together. As far as I know it is better to put them on top, to ensure mobile-first, but I couldn't find any info on that online, so you can disregard that if you want

FelixSchoberRubi (approved):

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+    .nav-menu {
+        display: none;
+        flex-direction: column;
+        position: absolute;
+        top: 100%;
+        left: 0;
+        right: 0;
+        background-color: #fbbf00;
+        padding: 1em;
+    }
```
IremA99, line 147: Maybe change padding to padding-bottom: 1em; to reduce the padding on top. Would look better but that just a personal opinion

IremA99: You could also remove the position:absolute to reduce the padding but again just a personal preference

IremA99 (approved):

RubeesEat/Pages/UserHomepage.cshtml
Ottorium: en puts a $ in front of the number. everywhere else we use en

RubeesEat/wwwroot/css/site.css
Ottorium: both dont do anything

RubeesEat/Pages/Shared/_Layout.cshtml
```diff
@@ -4,12 +4,39 @@
...
+        <li>
+            <a href="/SplitBill">Split a bill</a>
+        </li>
+        <li>
+            <a href="/LookBill">Show bills</a>
+        </li>
+    </ul>
+    <div class="logo-container">
+        <a href="/./">
+            <img src="/logo.png" alt="Logo" class="logo"/>
```
KuJo7, line 31: we have to use here a different rubicon logo since the background is already yellow.  We can select one from the public share (R:\Public\Marketing\Logos\RUBICON)

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+
+.hamburger span {
+    display: block;
+    width: 25px;
+    height: 3px;
+    background-color: black;
+    margin: 5px 0;
+    transition: all 0.3s ease;
+}
+
```
KuJo7, line 136:
```
.person {
    color: black;
}
```

KuJo7 (approved):

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+    justify-content: space-between;
+    align-items: center;
+    max-width: 1200px;
+    margin: 0 auto;
+}
+
+.logo-container {
+    display: flex;
+    align-items: center;
+}
```
a-ctor, line 96: As we have just one child, this does not do anything. We can remove the rule and the class on the element.

RubeesEat/Pages/Shared/_Layout.cshtml
```diff
@@ -4,12 +4,40 @@
...
+        <li>
+            <a href="/SplitBill">Split a bill</a>
+        </li>
+        <li>
+            <a href="/LookBill">Show bills</a>
+        </li>
+    </ul>
+    <div class="logo-container">
+        <a href="/./">
+            <img src="/better_logo.png" alt="Logo" class="logo"/>
```
a-ctor, line 32: `alt` text should be descriptive. Just `Logo` is probably not enough. Something like `Rubees Eat logo` might be better. In this case, in the `nav` bar it might be more apparent from context be if we can be specific we should be.

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+
+.nav-menu li {
+    margin-right: 1.5em;
+}
+
+.nav-menu li a {
+    color: black;
+    text-decoration: none;
+    font-weight: bold;
+}
```
a-ctor, line 118: Instead of adding a class of the list, we can also do `nav > ul`. This way, we don't have to introduce a class for each subelement of the root control (the `nav`).

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+.hamburger span {
+    display: block;
+    width: 25px;
+    height: 3px;
+    background-color: black;
+    margin: 5px 0;
+    transition: all 0.3s ease;
+}
+
+@media (max-width: 512px) {
```
a-ctor, line 137: Please but all the break points into CSS variables and use them instead. This allows us to change the breakpoints very easily.

a-ctor: Also CSS is usually developed mobile first. So all the main rules should apply to mobile. If the screen gets bigger we change stuff with media rules. So we would assert `min-width: xxx` instead and rewrite those rules.

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
...
+}
+
+.logo {
+    width: 50px;
+    height: 50px;
+    margin-left: auto;
+}
+
+nav {
+    background-color: #fbbf00;
```
a-ctor, line 74: Colors should be extracted to CSS variables. We should have a primary and secondary color. See issue description.

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
+/*source: https://www.joshwcomeau.com/css/custom-css-reset/*/
+*, *::before, *::after {
+    box-sizing: border-box;
+}
+
+* {
+    margin: 0;
+    font-family: "Metric Regular", Arial, Tahoma, sans-serif;
+    text-align: center;
```
a-ctor, line 9: Please no. This is not a good fix for "I want most of my stuff aligned to the middle". This rule applies to every element and thus makes styling very hard if you don't want `text-align: center`. It also disables inheritance completely which is not good.

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,172 @@
+/*source: https://www.joshwcomeau.com/css/custom-css-reset/*/
+*, *::before, *::after {
+    box-sizing: border-box;
+}
+
+* {
+    margin: 0;
+    font-family: "Metric Regular", Arial, Tahoma, sans-serif;
```
a-ctor, line 8: Please apply the `font-family` to the `body` element. Similar to the `text-align` comment this is not the right way as it disables inheritance.

RubeesEat/wwwroot/css/UserHomePage.css
```diff
@@ -0,0 +1,112 @@
+.positive-balance {
```
a-ctor, line 1: This CSS added in this file was not part of the issue. We don't want to change the content styling in this issue. As such, we are going to revert these changes. I would suggest starting from scratch and styling the page again as this has a few issues:

- Too many selectors reference specific elements on the page (`#list`, `#hello`, etc.). We should try and make reusable components that we can use on multiple pages.
- There are a lot of "need spacing here" changes. Ideally, we don't need to apply many margin/paddings to components individually as they should have good defaults
- Too many font size changes that should also be handled on a global scale. We don't want the sizes to be all over the place so there should be the same sizes regardless of which page the user is viewing.

RubeesEat/Pages/UserHomepage.cshtml
```diff
@@ -3,33 +3,35 @@
 @inject IBillRepository billRepository
 @inject IPersonRepository personRepository
 
-Hello, @personRepository.GetCurrentUser().FirstName!
+<h1 id="hello">Hello, @personRepository.GetCurrentUser().FirstName!</h1>
```
a-ctor, line 6: Revert the changes in this file (see other comment in the css)

a-ctor (changes requested):

RubeesEat/Pages/BillDetails.cshtml
```diff
@@ -17,7 +17,7 @@ Bill Details:
         <div class="date">@item.Date.ToString("D")</div>
         <div class="description">@item.Description</div>
         <div class="amount" style="display: inline;">Total amount: @item.EntryLines[0].Amount€</div>
-        <ul>
+        <ul id="list">
```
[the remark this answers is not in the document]
FelixSchoberRubi: When trying to fix this issue, we stumbled on the fact that all 3 css stylesheets are in the _Layout.cshtml file:
    <link rel="stylesheet" href="/css/site.css"/>
    <link rel="stylesheet" href="/css/UserHomePage.css"/>
    <link rel="stylesheet" href="/css/BillDetails.css"/>
    
   This does not make sense right? The id "list" is actually selected in the UserHomePage stylesheet, even though the ul with that id is in BillDetails.
   
  Should we change it, even though you didnt specifiy the change in the PR?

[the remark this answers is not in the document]
a-ctor: As discussed, we revert on of those and, in the future, we are also going to ensure that we don't do global CSS imports if they are not necessary.

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: I tried using CSS variables, but it didn't work for media queries. When I run it through a CSS validator, I get a parse error. Apparently it's not supported to use CSS variables in media queries: https://stackoverflow.com/questions/40722882/css-native-variables-not-working-in-media-queries

I will put the media queries at the top, after root, to be easily changeable

IremA99 (approved):

RubeesEat/wwwroot/css/site.css
a-ctor: Okay well, then we can leave them as is. I wouldn't move them too much as they should be close to the non media rules they semantically belong to.

RubeesEat/Pages/Index.cshtml
```diff
@@ -3,33 +3,35 @@
...
     var sign = isPositive ? "+" : "-";
     var amountAsString = sign + " " + amount.ToString("C", CultureInfo.CreateSpecificCulture("de-DE"));
+    var displayClass = isPositive ? "positive-balance" : "negative-balance";
 }
 
-<div id="balance" style="color: @displayColor">@amountAsString</div>
+<div id="balance" class="@displayClass">@amountAsString</div>
 
-<ul id="balanceChanges" style="list-style: none">
+<ul class = "entryLines">
```
a-ctor, line 18:
```suggestion
<ul class="entryLines">
```

RubeesEat/wwwroot/css/site.css
```diff
@@ -0,0 +1,128 @@
...
+    padding: 1em 2em;
+    position: sticky;
+    top: 0;
+    z-index: 1000;
+    display: flex;
+    justify-content: space-between;
+    align-items: center;
+}
+
+nav > ul {
```
a-ctor, line 47: On mobile, the elements are not centered correctly. On desktop, we want the elements next to the logo on the left side with the right side being empty.

a-ctor (changes requested):

PR #16, opened 2024-08-29
Feature/dseat 25 docker

Ottorium:
https://re-motion.atlassian.net/browse/DSEAT-25

### DIFF
--- a/RubeesEat/Dockerfile
+++ b/RubeesEat/Dockerfile
@@ -1,0 +1,19 @@
+# to build: docker build -t <name> .
+# to run: docker run -p <port(use port above 1024)>:8080 <name> .
+
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build-env
+WORKDIR /App
+ 
+# Copy everything
+COPY . ./
+# Restore as distinct layers
+RUN dotnet restore
+# Build and publish a release
+RUN dotnet publish -c Release -o out
+ 
+# Build runtime image
+FROM mcr.microsoft.com/dotnet/aspnet:8.0
+WORKDIR /App
+COPY --from=build-env /App/out .
+EXPOSE 8080
+ENTRYPOINT ["dotnet", "RubeesEat.dll"]


Ottorium:
do we need a .dockerignore file?

RubeesEat/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM mcr.microsoft.com/dotnet/sdk:8.0 AS build-env
+WORKDIR /App
+
```
a-ctor, line 3: Please add `EXPOSE 8080` to document that the image will listen to port 8080 TCP by default.

RubeesEat/RubeesEat.csproj
```diff
@@ -6,4 +6,11 @@
         <ImplicitUsings>enable</ImplicitUsings>
     </PropertyGroup>
 
+    <ItemGroup>
+      <Content Update="Dockerfile">
+        <ExcludeFromSingleFile>true</ExcludeFromSingleFile>
+        <CopyToPublishDirectory>PreserveNewest</CopyToPublishDirectory>
+      </Content>
+    </ItemGroup>
+
```
a-ctor, line 15: I don't think we need this. The `Dockerfile` is not necessary for the project output and thus does not need to be marked as `Content`.

RubeesEat/Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+RUN dotnet publish -c Release -o out
+ 
+# Build runtime image
+FROM mcr.microsoft.com/dotnet/aspnet:8.0
+WORKDIR /App
+COPY --from=build-env /App/out .
+ENTRYPOINT ["dotnet", "RubeesEat.dll"]
+
+# to build: docker build -t <name> .
+# to run: docker run -p <port(use port above 1024)>:8080 <name> .
```
a-ctor, line 18: Personal pref. but I would put that up to to make it the first thing you see when opening the file

a-ctor (changes requested):

RubeesEat/Dockerfile
FelixSchoberRubi: We tried adding a port with this comment. Then we had to set a port number in Docker Desktop to run it. But it didn't work

FelixSchoberRubi: Never mind, now it works.
 we didn't use 8080 before, I guess that was the issue

RubeesEat/RubeesEat.csproj
```diff
@@ -6,4 +6,5 @@
         <ImplicitUsings>enable</ImplicitUsings>
     </PropertyGroup>
 
+
```
a-ctor, line 9: ?

a-ctor (approved):

a-ctor:
> do we need a .dockerignore file?

No, we do the build in its own container and only put the build output in the image that we push. As such, the image should already only contain the application files and nothing else.

PR #20, opened 2024-09-06
Feature/dseat 24 styling for common web elements

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-24

No Peer review because task was done with everyone on the team

### DIFF
--- a/RubeesEat/Pages/SplitBill.cshtml
+++ b/RubeesEat/Pages/SplitBill.cshtml
@@ -10,7 +10,7 @@ Split bill:
     <label for="billAmount">Total Price:</label>
     <input id="billAmount" type="number" step=".01" name="billAmount" required="required"><br>
 
-    <span id="addedPeople"></span>
+    <label id="addedPeople"></label>
 
     <label for="billPeople">Add Person:</label>
     <select name="billPeople" id="billPeople">
--- a/RubeesEat/Pages/SplitBill.cshtml
+++ b/RubeesEat/Pages/SplitBill.cshtml
@@ -10,7 +10,7 @@ Split bill:
     <label for="billAmount">Total Price:</label>
     <input id="billAmount" type="number" step=".01" name="billAmount" required="required"><br>
 
-    <label id="addedPeople"></label>
+    <div id="addedPeople"></div>
 
     <label for="billPeople">Add Person:</label>
     <select name="billPeople" id="billPeople">
--- a/RubeesEat/wwwroot/css/site.css
+++ b/RubeesEat/wwwroot/css/site.css
@@ -289,11 +289,17 @@ input[type="button"].button-primary:focus {
     background-color: var(--positive-background-color);
 }
 
-.amount.negative-amount {
+.amount-description{
+    font-size: 2rem;
+    display: grid;
+    grid-template-columns: 25% 50% 25%;
+}
+
+.negative-amount {
     color: red;
 }
 
-.amount.positive-amount {
+.positive-amount {
     color: green;
 }
 
@@ -301,13 +307,20 @@ input[type="button"].button-primary:focus {
     transform: scale(1.02);
 }
 
+
+
 .clickBillDetails {
     text-decoration: none;
     color: var(--secondary-color);
-    display: flex;
-    justify-content: space-between;
-    align-items: center;
-    padding: 10px;
+}
+
+#balance {
+    background-color: var(--secondary-color);
+    font-size: 2rem;
+    width: 25%;
+    margin: auto;
+    border: 2px solid var(--primary-color);
+    border-radius: 5px;
 }
 
 @media (min-width: 512px)
--- a/RubeesEat/wwwroot/css/site.css
+++ b/RubeesEat/wwwroot/css/site.css
@@ -131,6 +131,15 @@ p, h1, h2, h3, h4, h5, h6 {
     font-weight: 300;
 }
 
+div
+{
+    overflow-wrap: break-word;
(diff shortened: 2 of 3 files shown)


RubeesEat/wwwroot/css/site.css
```diff
@@ -113,6 +110,193 @@ nav > ul > li:hover {
     transform: rotate(-45deg) translate(7px, -6px);
 }
 
+p, h1, h2, h3, h4, h5, h6 {
```
a-ctor, line 113: As you took a lot of stuff from skeleton css. Please add a notice up top that satisfies skeleton CSS's license.

RubeesEat/wwwroot/css/site.css
```diff
@@ -113,6 +110,193 @@ nav > ul > li:hover {
...
+h3 { font-size: 2.0rem; line-height: 1.3;  letter-spacing: -.1rem; }
+h4 { font-size: 1.5rem; line-height: 1.35; letter-spacing: -.08rem; }
+h5 { font-size: 1.0rem; line-height: 1.5;  letter-spacing: -.05rem; }
+h6 { font-size: 0.5rem; line-height: 1.6;  letter-spacing: 0; }
+
+/* Forms
+–––––––––––––––––––––––––––––––––––––––––––––––––– */
+input[type="number"],
+input[type="text"],
+select {
```
a-ctor, line 133: The `select` element has bug where the layout changes when the element is focused vs. not focused.

RubeesEat/wwwroot/css/site.css
```diff
@@ -113,6 +110,193 @@ nav > ul > li:hover {
...
+button.button-primary:focus,
+input[type="submit"].button-primary:focus,
+input[type="reset"].button-primary:focus,
+input[type="button"].button-primary:focus {
+    color: var(--secondary-color);
+    background-color: var(--primary-color-lighter);
+    border-color: var(--secondary-color); 
+}
+
+.entryLines {
```
a-ctor, line 252: The way the rules are formulated with `entryLines` as name suggests that it is not really reusable but should be moved into CSS specific to the site that is using the class.

If you want to, we can create a reusable "list" component but that should be more general and not have too specific styling for the one use case we have right now. So some of the rules might still need to be moved to a site specific CSS file.

Also remember that we can redesign the HTML of the list as well. We don't need to stay with an `<ul>` element and completely restyle it. For the kind of data we have here, a tabular display would be better in my opinion as it would make the amounts align vertically.

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -10,7 +10,7 @@ Split bill:
     <label for="billAmount">Total Price:</label>
     <input id="billAmount" type="number" step=".01" name="billAmount" required="required"><br>
 
-    <span id="addedPeople"></span>
+    <label id="addedPeople"></label>
```
a-ctor, line 13: `<label>` should only be used together with another element that is referenced using the `for` attribute. Here we just want a container for elements so we should use `<div>` instead.

a-ctor (changes requested):

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: We use the entrylines again in "ShowBills", and also in an individual Bill (but without the "positive/negative amount class")
image

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: According to this, we need to add the MIT License, so I will do that 
https://github.com/dhg/Skeleton?tab=readme-ov-file#license

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: Regarding the tabular display: In DSEAT-32 (https://github.com/re-motion/students-RubeesEat/pull/22) I used flexbox and margin:auto, flex-grow 1, etc. to make the entrylines display like a table.
I did it there rather than here because it was more of a issue with the master layout

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: I don't see the bug when I try it myself. Weirdly enough, when I try to set the border and outline in the normal select selector, and then only change the color when focused (so like this):

select {
height: 38px;
padding: 6px 10px; /* The 6px vertically centers text on FF, ignored by Webkit */
background-color: white;
border-radius: 4px;
border: 1px solid;
outline: 0;
max-width: 200px;
}

input[type="number"]:focus,
input[type="text"]:focus,
select:focus {
border: var(--primary-color);
}

THEN i suddenly get a bug where the text of the select element starts moving around when I click on it and off it, even though I only want to change the color. But before I didn't have that bug

FelixSchoberRubi: Turns out it was already fixed when discussing it during standup

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: It turns out that if we use divs instead of li, the divs underneath do not inherit the background color. I'm leaving it as a li for now.
Now we changed it to use grid layout

IremA99 (approved):

KuJo7 (approved):

a-ctor (approved):

PR #21, opened 2024-09-09
Feature/dseat 31 improve design of individual bill page

IremA99:
https://re-motion.atlassian.net/browse/DSEAT-31

### DIFF
--- a/RubeesEat/wwwroot/css/BillDetails.css
+++ b/RubeesEat/wwwroot/css/BillDetails.css
@@ -1,0 +1,69 @@
+div.bill-summary {
+    padding: 1.5rem;
+    background-color: var(--secondary-color);
+    border: 1px solid var(--primary-color);
+    border-radius: 8px;
+    max-width: 400px;
+    margin: 1rem auto;
+    text-align: center;
+}
+
+div.bill-summary > .personWhoPaid {
+    font-weight: bold;
+    font-size: 1.2em;
+    margin-bottom: 10px;
+}
+
+div.bill{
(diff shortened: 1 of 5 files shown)


RubeesEat/wwwroot/css/site.css
```diff
@@ -113,6 +110,193 @@ nav > ul > li:hover {
...
+    color: red;
+}
+
+.amount.positive-amount {
+    color: green;
+}
+
+.entryLines li:hover {
+    transform: scale(1.02);
+}
```
KuJo7, line 289: I think this hover effect should only apply to the entrylines in home and not on billdetails too since this is not clickable

KuJo7 (approved):

a-ctor (changes requested):
Please re-request once the other PR (which this one depends on) is merged. Please also rebase this branch to the top before re-requesting

RubeesEat/wwwroot/css/BillDetails.css
```diff
@@ -0,0 +1,59 @@
...
+    padding: 1.5rem;
+    background-color: var(--secondary-color);
+    border: 1px solid var(--primary-color);
+    border-radius: 8px;
+    max-width: 400px;
+    margin: 1rem auto;
+    text-align: center;
+}
+
+.billDetails{
```
KuJo7, line 11: I think we should get rid of the hover effect, because this indicates a button which is actually not the case

a-ctor (approved):

PR #22, opened 2024-09-09
Feature/dseat 32 improve master layout

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-32

### DIFF
--- a/RubeesEat/wwwroot/css/site.css
+++ b/RubeesEat/wwwroot/css/site.css
@@ -271,8 +271,8 @@ input[type="button"].button-primary:focus {
 }
 
 .entryLines {
-    --positive-background-color: #FFC1C1;
-    --negative-background-color: #E7FFE7;
+    --positive-background-color: #E7FFE7;
+    --negative-background-color: #FFC1C1;
 }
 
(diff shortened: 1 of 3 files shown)


RubeesEat/wwwroot/css/site.css
```diff
@@ -250,12 +250,18 @@ input[type="button"].button-primary:focus {
...
+    align-items: center;
 }
 
 .entryLines li {
+    max-width: 800px;
+    width: 100%;
+    word-wrap: break-word;
     margin: 10px 10%;
     padding: 10px;
     background-color: white;
```
IremA99, line 267: Why is the background color set to white here?

IremA99 (approved):

RubeesEat/wwwroot/css/site.css
FelixSchoberRubi: This is for a general Entryline, that does not have a background color for positive/negative amount. This is for looking at an individual bill (where there is no positive/negative amount)

PR #23, opened 2024-09-09
Feature/dseat 30 add show bills functionality

IremA99:
https://re-motion.atlassian.net/browse/DSEAT-30

### DIFF
--- a/RubeesEat/Model/InMemoryBillRepository.cs
+++ b/RubeesEat/Model/InMemoryBillRepository.cs
@@ -1,3 +1,4 @@
+using System.Collections.Immutable;
 namespace RubeesEat.Model;
 
 public class InMemoryBillRepository : IBillRepository
@@ -31,32 +32,46 @@ public class InMemoryBillRepository : IBillRepository
         ArgumentNullException.ThrowIfNull(bill);
         _bills.Add(bill.Id, bill);
     }
+        
+    public PaginatedView<BalanceChange> GetRecentBalanceChanges(Person currentUser, int page, int pageSize)
+    {
+        var skip = (page - 1) * pageSize;
+
+        var billsOfUser = _bills.Values
+                                .Where(b => b.EntryLines.Any(e => e.Person.Id == currentUser.Id))
+                                .ToList();
+
+        int totalPages = (int)Math.Ceiling(billsOfUser.Count / (double)pageSize);
+
+        var balanceChanges = billsOfUser
+                             .OrderByDescending(b => b.Date)
+                             .Skip(skip)
+                             .Take(pageSize)
+                             .Select(b => GetBalanceChangesForUser(b, currentUser))
+                             .ToImmutableArray();
+        
+        var billsPaginated = new PaginatedView<BalanceChange>(
+            balanceChanges,
+            page,
+            totalPages
+        );
+
+        return billsPaginated;
+        
+        static BalanceChange GetBalanceChangesForUser(Bill bill, Person currentUser)
+        {
+            var sum = bill.EntryLines
+                          .Where(entryLine => entryLine.Person.Id == currentUser.Id)
+                          .Sum(entryLine => entryLine.Amount);
+            return new BalanceChange(sum, bill.Date, bill.Description, bill.Id);
+        }
+    }
 
     public IReadOnlyList<Bill> GetAllForUser(Person user)
     {
         return _bills.Values
                      .Where(e => e.EntryLines.Any(f => f.Person == user))
                      .ToList();
-    }
-
-    public IReadOnlyList<BalanceChange> GetRecentBalanceChanges(Person user, int amount)
-    {
-        var billsWithUser = GetAllForUser(user).OrderByDescending(bill => bill.Date);
-        var balanceChanges = new List<BalanceChange>();
-        int counter = 0;
-        foreach (var bill in billsWithUser)
-        {
-            if (counter >= amount)
-                break;
-            var sum = bill.EntryLines
-                          .Where(entryLine => entryLine.Person.Id == user.Id)
-                          .Sum(entryLine => entryLine.Amount);
-
-            balanceChanges.Add(new BalanceChange(sum, bill.Date, bill.Description, bill.Id));
-            counter++;
-        }
-
-        return balanceChanges;
     }
 
     public decimal GetBalance(Person user)
--- a/RubeesEat/Pages/Shared/_Layout.cshtml
+++ b/RubeesEat/Pages/Shared/_Layout.cshtml
@@ -26,7 +26,7 @@
             <a href="/SplitBill">Split a bill</a>
         </li>
         <li>
-            <a href="/LookBill">Show bills</a>
+            <a href="/ShowBills">Show bills</a>
         </li>
     </ul>
     <button class="hamburger">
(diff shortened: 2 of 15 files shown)


RubeesEat/Pages/ShowBills.cshtml
```diff
@@ -0,0 +1,62 @@
...
+@using System.Globalization
+@inject IBillRepository billRepository
+@inject IPersonRepository personRepository
+
+<h1>Your Bills</h1>
+
+@{
+    var currentUser = personRepository.GetCurrentUser();
+    var pageSize = 3;
+    int pageNumber;
```
FelixSchoberRubi, line 11: change to pageIndex, to be consistent with the property in paginatedList class

FelixSchoberRubi (changes requested):

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -28,6 +28,14 @@ public IReadOnlyList<Bill> GetAll()
             : null;
     }
 
+    public BalanceChange GetBalanceChangesForUser(List<BalanceChange> balanceChanges, Bill bill, Person currentUser)
```
FelixSchoberRubi, line 31: Parameter List<BalanceChange> balanceChanges is not used in function and can be removed

FelixSchoberRubi (approved):

a-ctor (changes requested):

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -26,6 +26,39 @@ public IReadOnlyList<Bill> GetAll()
...
+
+            return balanceChanges.Skip((int)skip)
+                                 .Take((int)pageSize)
+                                 .Select(b => GetBalanceChangesForUser(b, user))
+                                 .ToList();
+        }
+
+        return new List<BalanceChange>();
+    }
+
```
FelixSchoberRubi, line 61: Maybe there is a way to simplify this. 

1. We can use our helper function to reduce duplication
2. Currently, this code gets all balance changes from a user, then makes another list out of that, based on if we want a certain amount or balance changes for the Show Bills page.

We can simplify by removing the amount parameter. Instead, if we want recent balance changes, we set pageIndex to 1 and pageSize to the amount of pages we want. 

So we can leave this function as we had it previously

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -38,27 +71,7 @@ public IReadOnlyList<Bill> GetAllForUser(Person user)
...
-                          .Sum(entryLine => entryLine.Amount);
-
-            balanceChanges.Add(new BalanceChange(sum, bill.Date, bill.Description, bill.Id));
-            counter++;
-        }
-
-        return balanceChanges;
-    }
-
+
```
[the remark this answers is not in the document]
FelixSchoberRubi: I mean that there is a whitespace in line 74. Github somehow doesnt mark the lines properly

RubeesEat.IntegrationTests/SeleniumTests/UserHomepageSeleniumTests.cs
```diff
@@ -22,14 +22,6 @@ public void GetDefaultBalanceChanges()
...
         Assert.That(balanceChanges[2].Date, Is.EqualTo("Sunday, April 28, 2024"));
         Assert.That(balanceChanges[2].Description, Is.EqualTo("Lunch with default user in 2024"));
-
-        Assert.That(balanceChanges[3].Amount, Is.EqualTo("-800,00 \u20ac"));
-        Assert.That(balanceChanges[3].Date, Is.EqualTo("Friday, April 28, 2023"));
-        Assert.That(balanceChanges[3].Description, Is.EqualTo("Lunch with default user in 2023"));
-
-        Assert.That(balanceChanges[4].Amount, Is.EqualTo("-800,00 \u20ac"));
-        Assert.That(balanceChanges[4].Date, Is.EqualTo("Thursday, April 28, 2022"));
-        Assert.That(balanceChanges[4].Description, Is.EqualTo("Lunch with default user 2022"));
```
FelixSchoberRubi, line 32: For context: now that we have a page for showing bills, the front page only shows the 3 most recent bills, instead of 5

FelixSchoberRubi (approved):

KuJo7 (approved):

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -39,31 +66,19 @@ public IReadOnlyList<Bill> GetAllForUser(Person user)
...
-
     public decimal GetBalance(Person user)
     {
         return _bills.Values
                      .SelectMany(e => e.EntryLines)
                      .Where(e => e.Person == user)
                      .Sum(e => e.Amount);
     }
+
+    private BalanceChange GetBalanceChangesForUser(Bill bill, Person currentUser)
```
a-ctor, line 77: Please convert this to a static local method at the bottom of the method it is used in.

a-ctor (changes requested):

RubeesEat/Model/InMemoryBillRepository.cs
```diff
@@ -51,12 +57,14 @@ public PaginatedView<BalanceChange> GetBalanceChangesForPage(Person currentUser,
...
 
         return billsPaginated;
-    }
-
-    public void Add(Bill bill)
-    {
-        ArgumentNullException.ThrowIfNull(bill);
-        _bills.Add(bill.Id, bill);
+        
+        BalanceChange GetBalanceChangesForUser(Bill bill, Person currentUser)
```
FelixSchoberRubi, line 61: needs to be static according to PR review

FelixSchoberRubi (approved):

RubeesEat/Pages/Shared/_Layout.cshtml
```diff
@@ -22,7 +22,7 @@
             <a href="/SplitBill">Split a bill</a>
         </li>
         <li>
-            <a href="/LookBill">Show bills</a>
```
a-ctor, line 25: `/ShowBills/1` looks like we try to look at a bill with the ID 1 but it indicates the page instead. I think we should do two things here:

1. Use the query to provide the page (e.g. `/ShowBills?page=1`) instead of the route.
2. Add a default to the value so that we don't have to provide it here. This way, we can use `/ShowBills` as link and it will automatically use the first page

a-ctor (changes requested):

RubeesEat/Pages/ShowBills.cshtml
```diff
@@ -11,11 +11,13 @@
...
 
-    int pageIndex;
-    if (!Int32.TryParse(RouteData.Values["pageNumber"].ToString(), out pageIndex))
-        pageIndex = 1;
+    int page = 1;
     
-    var billsPaginated = billRepository.GetBalanceChanges(currentUser, pageIndex, pageSize);
+    if (!Int32.TryParse(Request.Query["page"], out page) || page <= 0)
+    {
+        page = 1;
```
FelixSchoberRubi, line 18: page = 1 happens twice

FelixSchoberRubi (approved):

a-ctor (approved):

PR #29, opened 2024-09-17
DSEAT-39 Added selenium test for removing people from SplitBill page

IremA99:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/

### DIFF
--- a/RubeesEat.IntegrationTests/PageObjects/SpiltBillPageObject.cs
+++ b/RubeesEat.IntegrationTests/PageObjects/SpiltBillPageObject.cs
@@ -45,4 +45,11 @@ public class SplitBillPageObject : PageObject
                .Select(e => new PersonAmountControlObject(this, e))
                .ToArray();
     }
+
+    public List<string> GetSelection()
+    {
+        var selectElement = FindElement(By.Id("billPeople"));
+        var options = selectElement.FindElements(By.TagName("option"));
+        return options.Select(e => e.Text).ToList();
+    }
 }
--- a/RubeesEat.IntegrationTests/SeleniumTests/SplitBillSeleniumTests.cs
+++ b/RubeesEat.IntegrationTests/SeleniumTests/SplitBillSeleniumTests.cs
@@ -98,5 +98,26 @@ public class SplitBillSeleniumTests() : SeleniumIntegrationTestBase("SplitBill")
         
         Assert.That(page.GetErrorMessage(), Is.EqualTo("The sum of individual amounts doesn't match the total bill amount."));
     }
+
+    [Test]
+    public void AddAndRemovePerson()
+    {
+        var page = Start<SplitBillPageObject>();
+        
+        var selectionsStart = page.GetSelection();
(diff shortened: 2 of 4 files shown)


FelixSchoberRubi (approved):

RubeesEat.IntegrationTests/SeleniumTests/SplitBillSeleniumTests.cs
```diff
@@ -98,5 +98,20 @@ public void WrongAmountPerPerson()
...
+    {
+        var page = Start<SplitBillPageObject>();
+        
+        var person = page.ClickAddPerson();
+        
+        Assert.That(page.GetPersonAmountAndNames().Length, Is.EqualTo(1));
+        
+        person.ClickRemoveButton();
+        
+        Assert.That(page.GetPersonAmountAndNames().Length, Is.EqualTo(0));
```
a-ctor, line 113: I think it would be important here to assert that the person is now available again in the list of people to choose. So we should assert that when we add a person it is no longer in the list. And if we remove it, it is back in the list of items to choose from.

a-ctor (changes requested):
See other comment. I pressed too quickly.

RubeesEat.IntegrationTests/SeleniumTests/SplitBillSeleniumTests.cs
```diff
@@ -104,14 +104,17 @@ public void AddAndRemovePerson()
...
         Assert.That(page.GetPersonAmountAndNames().Length, Is.EqualTo(1));
+        Assert.That(page.GetSelection().Count < selectionsStart.Count());
         
         person.ClickRemoveButton();
         
         Assert.That(page.GetPersonAmountAndNames().Length, Is.EqualTo(0));
-
+        Assert.That(page.GetSelection().Count, Is.EqualTo(selectionsStart.Count()));
     }
 }
```
FelixSchoberRubi, line 119: I think we should assert if that specific person is removed from the options once we add it, and once we remove they are back on the options. So not only assert the length, but also if it was the right person

FelixSchoberRubi (changes requested):

KuJo7 (approved):

FelixSchoberRubi (approved):

RubeesEat.IntegrationTests/PageObjects/SpiltBillPageObject.cs
```diff
@@ -45,4 +45,16 @@ public PersonAmountControlObject[] GetPersonAmounts()
...
+    public List<string> GetSelection()
+    {
+        var selectElement = FindElement(By.Id("billPeople"));
+        var options = selectElement.FindElements(By.TagName("option"));
+        List<string> selection = new List<string>();
+        foreach (var option in options)
+        {
+            selection.Add(option.Text);
+        }
+        return selection;
```
a-ctor, line 58: We can simplify this using LINQ

a-ctor (approved):

PR #30, opened 2024-09-17
DSEAT-38 Taghelpers for tests

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-38


RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
```diff
@@ -27,6 +29,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
             return Me.FindElements(by);
         }
         
+        public IWebElement FindElement(string t_id)
+        {
+            ReadOnlyCollection<IWebElement> elements = Me.FindElements(By.CssSelector($"[data-test-id='{t_id}']"));
+            if (elements.Count == 0)
+            {
```
KuJo7, line 36: we could remove the braces for single line ifs

RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
```diff
@@ -27,6 +29,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+                case "ClickAndWaitUntilStale":
+                    element.ClickAndWaitUntilStale();
+                    break;
+                case "Click":
+                    element.Click();
+                    break;
+                default:
+                    throw new UnsupportedClickBehavior(clickBehavior, action);
+            }
```
KuJo7, line 70: By using switch expressions, the code becomes more concise and readable
```suggestion
            clickBehavior switch
            {
                "ClickAndWaitUntilStale" => element.ClickAndWaitUntilStale(),
                "Click" => element.Click(),
                _ => throw new UnsupportedClickBehavior(clickBehavior, action);
            };
```

RubeesEat.IntegrationTests/SeleniumIntegrationTestBase.cs
```diff
@@ -31,6 +31,9 @@ public void SetUp()
         _webHost.Start();
 
         WebTestHelper = new WebTestHelper();
+        
+        Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", "Test");
+
     }
```
IremA99, line 37: I know that was me but please revert the whitespace at the end

RubeesEat.IntegrationTests/WebTesting/PageObject.cs
```diff
@@ -26,6 +28,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+        switch (clickBehavior)
+        {
+            case "ClickAndWaitUntilStale":
+                element.ClickAndWaitUntilStale();
+                break;
+            case "Click":
+                element.Click();
+                break;
+            default:
+                throw new UnsupportedClickBehavior(clickBehavior, action);
```
KuJo7, line 69: see comment above

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,29 @@
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace RubeesEat.TagHelpers;
+
+[HtmlTargetElement(Attributes = "t-action")]
+public class WebTestActionTagHelper : TagHelper
+{ 
+    public string TAction { get; set; }
+    public string TClickBehavior { get; set; }
+    private readonly IWebHostEnvironment _env;
```
KuJo7, line 10: private field should be written before public properties
```suggestion
    private readonly IWebHostEnvironment _env;
    
    public string TAction { get; set; }
    public string TClickBehavior { get; set; }
```

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+    }
+    
+    public override void Process(TagHelperContext context, TagHelperOutput output)
+    {
+        if (!(_env.IsProduction()))
+        {
+            output.Attributes.RemoveAll("t-action");
+            output.Attributes.RemoveAll("t-click-type");
+            output.Attributes.Add("data-test-action", TAction);
+            if (TClickBehavior == null)
```
KuJo7, line 24:
```suggestion

            if (TClickBehavior == null)
```

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,29 @@
...
+                TClickBehavior = "ClickAndWaitUntilStale";
+            output.Attributes.Add("data-test-click-behavior", TClickBehavior);
```
KuJo7, line 26:
```suggestion

            output.Attributes.Add("data-test-click-behavior", TClickBehavior);
```

RubeesEat/TagHelpers/WebTestIdTagHelper.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace RubeesEat.TagHelpers;
+
+[HtmlTargetElement(Attributes = "t-id")]
+public class WebTestIdTagHelper : TagHelper
+{ 
+    public string TId { get; set; }
+    private readonly IWebHostEnvironment _env;
```
KuJo7, line 9: see comment above

KuJo7 (approved):

RubeesEat.IntegrationTests/SeleniumIntegrationTestBase.cs
```diff
@@ -31,6 +31,9 @@ public void SetUp()
         _webHost.Start();
 
         WebTestHelper = new WebTestHelper();
+        
+        Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", "Test");
+
```
IremA99, line 36:
```suggestion

```

RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
```diff
@@ -27,6 +29,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+            ReadOnlyCollection<IWebElement> elements = Me.FindElements(By.CssSelector($"[data-test-id='{t_id}']"));
+            if (elements.Count == 0)
+            {
+                throw new NoSuchElementException($"Element with id: {t_id} was not found.");
+            }
+
+            if (elements.Count > 1)
+            {
+                throw new TooManyElementsException();
+            }
```
IremA99, line 43:
```suggestion
            }

```

RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
```diff
@@ -27,6 +29,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+            ReadOnlyCollection<IWebElement> elements = Driver.FindElements(By.CssSelector($"[data-test-action='{action}']"));
+            if (elements.Count == 0)
+            {
+                throw new NoSuchElementException($"Element with id: {action} was not found.");
+            }
+            if (elements.Count > 1)
+            {
+                throw new TooManyElementsException();
+            }
+            var element = elements.First();
```
IremA99, line 58:
```suggestion

            var element = elements.First();
```

RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
FelixSchoberRubi: Unfortunately, the code has an error if i use that. the ; on line 64 is wrong, and if i remove it, it is still an error because switch expressions only work with expressions, not statements.

RubeesEat.IntegrationTests/WebTesting/PageObject.cs
FelixSchoberRubi: see above

RubeesEat.IntegrationTests/WebTesting/PageObject.cs
```diff
@@ -26,6 +28,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+        if (elements.Count == 0)
+        {
+            throw new NoSuchElementException($"Element with id: {data_test_id} was not found.");
+        }
+
+        if (elements.Count > 1)
+        {
+            throw new TooManyElementsException();
+        }
+        return elements.First();
```
IremA99, line 44: If you are going to change the code above to single line ifs (as joel suggested) you should do it here as well

RubeesEat.IntegrationTests/WebTesting/PageObject.cs
```diff
@@ -26,6 +28,47 @@ public IReadOnlyCollection<IWebElement> FindElements(By by)
...
+    {
+        ReadOnlyCollection<IWebElement> elements = Driver.FindElements(By.CssSelector($"[data-test-action='{action}']"));
+        if (elements.Count == 0)
+        {
+            throw new NoSuchElementException($"Element with id: {action} was not found.");
+        }
+        if (elements.Count > 1)
+        {
+            throw new TooManyElementsException();
+        }
```
IremA99, line 57: Same as above, single line ifs

IremA99 (changes requested):

IremA99 (approved):

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+namespace RubeesEat.TagHelpers;
+
+[HtmlTargetElement(Attributes = "t-action")]
+public class WebTestActionTagHelper : TagHelper
+{ 
+    private readonly IWebHostEnvironment _env;
+    
+    public string TAction { get; set; }
+    public string TClickBehavior { get; set; }
```
a-ctor, line 11: There is an attribute to change the name of the HTML attribute that the property binds to. Please use that so that we can use normal names for our properties. something like `[HtmlAttributeName("t-action")] public string Action { get; set; }` (Also do this in the other file please)

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        _env = env;
+    }
+    
+    public override void Process(TagHelperContext context, TagHelperOutput output)
+    {
+        if (!(_env.IsProduction()))
+        {
+            output.Attributes.RemoveAll("t-action");
+            output.Attributes.RemoveAll("t-click-type");
```
a-ctor, line 23: I am not sure that this is necessary (Also check the other file please)

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,32 @@
...
+
+[HtmlTargetElement(Attributes = "t-action")]
+public class WebTestActionTagHelper : TagHelper
+{ 
+    private readonly IWebHostEnvironment _env;
+    
+    public string TAction { get; set; }
+    public string TClickBehavior { get; set; }
+
+    public WebTestActionTagHelper(IWebHostEnvironment env)
```
a-ctor, line 13: If possible we should not use abbreviations for names -> `environment` (only exception is LINQ methods where I just default to `e` :p)

Please change in the other tag helper as well

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -19,9 +22,10 @@ Split bill:
             <option value="@person.FirstName @person.LastName" data-id="@person.Id">@person.FirstName @person.LastName </option>
         }
     </select>
-    <input id="addPersonButton" type="button" onclick="AddPerson()" value="Add Person"><br/>
+    <input id="addPersonButton" t-action="addPersonButton" t-click-behavior="Click" type="button" onclick="AddPerson()" value="Add Person"></input><br/>
 
     <button id="splitBillButton" type="button" onclick="validate()">Split the bill</button><br>
     <text id="placeForErrorMessage" style="color: red;"></text>
 </form>
+<script type="text/javascript">const isProduction = @env.IsProduction().ToString().ToLower();</script>
```
a-ctor, line 30: This will only apply the change for this one page, so if we want to have other pages with JS as well, this won't work. I would suggest moving this to the layout and inverting the variable. So we have a variable `isTest` which we set if the environment is `Test`, otherwise no JS is injected.

RubeesEat.IntegrationTests/SeleniumTests/SplitBillSeleniumTests.cs
```diff
@@ -9,7 +9,7 @@ public void AddPerson()
     {
         var page = Start<SplitBillPageObject>();
         
-        page.ClickAddPerson();
```
a-ctor, line 12: With the page object pattern all the finding stuff should be located in the page object and not added in the test. Previously, we called `ClickAddPerson` and the method on the page object decided what to do. Now we supply the action directly, which is not what we want. Please move the `InvokeAction` into the method of the page object.

RubeesEat.IntegrationTests/WebTesting/ControlObject.cs
```diff
@@ -1,5 +1,8 @@
 ﻿using OpenQA.Selenium;
 using OpenQA.Selenium.Chrome;
+using System.Collections.ObjectModel;
+using RubeesEat.Exceptions;
+using RubeesEat.IntegrationTests.PageObjects;
```
a-ctor, line 5: We probably don't need all of them

RubeesEat.IntegrationTests/WebTesting/PageObject.cs
```diff
@@ -1,5 +1,8 @@
 ﻿using OpenQA.Selenium;
 using OpenQA.Selenium.Chrome;
+using System.Collections.ObjectModel;
+using RubeesEat.Exceptions;
+using RubeesEat.IntegrationTests.PageObjects;
```
a-ctor, line 5: We probably don't need all of them

RubeesEat.IntegrationTests/WebTesting/WebTestElementFinder.cs
```diff
@@ -0,0 +1,42 @@
+using System.Collections.ObjectModel;
+using OpenQA.Selenium;
+using RubeesEat.Exceptions;
+using RubeesEat.IntegrationTests.WebTesting;
+
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public static class WebTestElementFinder
+{
+    public static IWebElement FindElement(ISearchContext context, string id)
```
a-ctor, line 10: Nice finding `ISearchContext` :)

RubeesEat.IntegrationTests/WebTesting/WebTestElementFinder.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        ReadOnlyCollection<IWebElement> elements = context.FindElements(By.CssSelector($"[data-test-id='{id}']"));
+        if (elements.Count == 0)
+            throw new NoSuchElementException($"Element with id: {id} was not found.");
+
+        if (elements.Count > 1)
+            throw new TooManyElementsException($"Element with id: {id} returned more than one element.");
+        return elements.First();
```
a-ctor, line 18:
```suggestion
        if (elements.Count > 1)
            throw new TooManyElementsException($"Element with id: {id} returned more than one element.");

        return elements.First();
```

RubeesEat.IntegrationTests/WebTesting/WebTestElementFinder.cs
```diff
@@ -0,0 +1,42 @@
...
+    
+    public static void InvokeAction(ISearchContext context, string action)
+    {
+        ReadOnlyCollection<IWebElement> elements = context.FindElements(By.CssSelector($"[data-test-action='{action}']"));
+        if (elements.Count == 0)
+            throw new NoSuchElementException($"Element with id: {action} was not found.");
+        if (elements.Count > 1)
+            throw new TooManyElementsException($"Element with id: {action} returned more than one element.");
+        var element = elements.First();
+        var clickBehavior = element.GetAttribute("data-test-click-behavior");
```
a-ctor, line 29:
```suggestion
        if (elements.Count > 1)
            throw new TooManyElementsException($"Element with id: {action} returned more than one element.");

        var element = elements.First();
        var clickBehavior = element.GetAttribute("data-test-click-behavior");
```

RubeesEat.IntegrationTests/WebTesting/WebTestElementFinder.cs
```diff
@@ -0,0 +1,42 @@
...
+
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public static class WebTestElementFinder
+{
+    public static IWebElement FindElement(ISearchContext context, string id)
+    {
+        ReadOnlyCollection<IWebElement> elements = context.FindElements(By.CssSelector($"[data-test-id='{id}']"));
+        if (elements.Count == 0)
+            throw new NoSuchElementException($"Element with id: {id} was not found.");
```
a-ctor, line 14: I would emphasize the fact that we are looking for a "test id" and not an HTML id. I would also suggest putting the parameter in single quotes to make the id more apparent in the output.

Please fix this in the exceptions below as well. For the action stuff we should also emphasize that we are missing a "test action"

RubeesEat/Pages/SplitBill.cshtml
```diff
@@ -1,11 +1,13 @@
 @page
 @inject IPersonRepository PersonRepository
+@addTagHelper *, Microsoft.AspNetCore.Mvc.TagHelpers
+@addTagHelper *, RubeesEat
```
a-ctor, line 4: We shouldn't need these as we have imports in the layout already?

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace RubeesEat.TagHelpers;
+
+[HtmlTargetElement(Attributes = "t-action")]
+public class WebTestActionTagHelper : TagHelper
+{ 
+    
+    private readonly IWebHostEnvironment _environment;
+    [HtmlAttributeName("t-action")] public string Action { get; set; }
+    [HtmlAttributeName("t-click-behavior")] public string ClickBehavior { get; set; }
```
a-ctor, line 11: Personaly preference but I like placing the attributes on the lines before the property to not inflate line length and make multiple attributes more readable.

```suggestion
    private readonly IWebHostEnvironment _environment;

    [HtmlAttributeName("t-action")]
    public string Action { get; set; }

    [HtmlAttributeName("t-click-behavior")]
    public string ClickBehavior { get; set; }
```

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+
+    public WebTestActionTagHelper(IWebHostEnvironment environment)
+    {
+        _environment = environment;
+    }
+    
+    public override void Process(TagHelperContext context, TagHelperOutput output)
+    {
+        if (!(_environment.IsProduction()))
```
a-ctor, line 20: Also personal preference: Inverted conditions are always harder to read than non inverted. I would therefore prefer `if (_environment.IsProduction()) return;`. This makes it clear that production does nothing as the method is exited and we don't have an extra `else` block with indent.

Another benefit of this "early exit" strategy compared to `if-else` is that I don't have to check if there is any code after the `if`. If I read if condition, I still have to check if there is something after the if which might happen in production. With the return variant, both lines are side by side making it easy to see that nothing else happens.

RubeesEat/TagHelpers/WebTestActionTagHelper.cs
```diff
@@ -0,0 +1,30 @@
...
+        {
+            output.Attributes.Add("data-test-action", Action);
+            
+            if (ClickBehavior == null)
+                ClickBehavior = "ClickAndWaitUntilStale";
```
a-ctor, line 25: We can assign the property above to the default value. This way, we don't need to check for `null` here and can simplify the logic.

RubeesEat/TagHelpers/WebTestIdTagHelper.cs
```diff
@@ -0,0 +1,24 @@
+using Microsoft.AspNetCore.Razor.TagHelpers;
+
+namespace RubeesEat.TagHelpers;
+
+[HtmlTargetElement(Attributes = "t-id")]
+public class WebTestIdTagHelper : TagHelper
```
a-ctor, line 6: See the comments in the other tag helper because they apply here as well.

RubeesEat/wwwroot/js/SplitBill.js
```diff
@@ -23,6 +23,9 @@ function AddPerson() {
     amountElement.name = "amount" + counter;
     amountElement.type = "number";
     amountElement.step = ".01";
+    if (isTest) {
+        amountElement.setAttribute("data-test-id", "input");
+    }
```
a-ctor, line 28: Either no braces or the opening brace on the next line ;)

a-ctor (changes requested):

RubeesEat/Pages/SplitBill.cshtml
IremA99: It doesn't work if i remove them. Also we don't have the imports in the layout. Should we move them there or is it okay to leave them here?

a-ctor (approved):

PR #32, opened 2024-09-19
DSEAT-21 MariaDB implementation

KuJo7:
https://re-motion.atlassian.net/browse/DSEAT-21


RubeesEat/Model/InMemory/InMemoryPersonRepository.cs
```diff
@@ -33,27 +32,15 @@ public void Add(Person person)
...
-        if (!_persons.ContainsKey(CurrentUser))
-            throw new ArgumentException("default User is not in the dictionary");
-        return _persons[CurrentUser];
+        return _persons.TryGetValue(CurrentUser, out var person) 
+            ? person 
+            : throw new ArgumentException("default User is not in the dictionary");
     }
 
-    public Person? GetById(Guid id)
+    public Person? GetById(Guid guid)
```
a-ctor, line 40: I would keep `id` as name.

RubeesEat/Model/Person.cs
```diff
@@ -23,8 +23,38 @@ public Person(Guid id, string firstName, string lastName)
...
     {
         return words.All(s => Regex.IsMatch(s, @"^[a-zA-Z]+$"));
     }
+    
+    public override bool Equals(object? obj)
+    {
+        if (obj is not Person other)
+            return false;
+
+        return Id == other.Id;
```
a-ctor, line 36: Now we have two different equality implementations for domain objects. One where we compare ID + properties (see `Bill`) and this one where we compare just the ID. We should decide on one implementation for domain objects and use it everywhere.

Using `IEqualityComparer<T>` we can implemented different equality comparison methods outside of the domain object class. I would suggest removing the equality from the domain objects entirely. This will revert them to reference equality. We can then implement `IEqualityComparer<T>` to have different equality contracts. So we would have one for the ID equality and one for full equality. Try to find a way that we don't have to write the ID equality comparer for all domain object separately but that we can reuse the same class for all domain objects.

RubeesEat/Startup.cs
```diff
@@ -16,51 +17,18 @@ public void ConfigureServices(IServiceCollection services)
...
-            new EntryLine(persons[0], -800),
-            new EntryLine(persons[3], -3200)
-        ]));
-        billRepository.Add(new Bill(Guid.NewGuid(), new DateTime(2024, 4, 28), "Lunch with default user in 2024",
-        [
-            new EntryLine(persons[1], 1_000),
-            new EntryLine(persons[0], -800),
-            new EntryLine(persons[3], -200)
-        ]));
+        var connectionString = $"Server=localhost;Database=RubeesEatDB;UID={username};PWD={password};";
```
a-ctor, line 23: We might want to change the connection string as a whole. So I would suggest passing the whole connection string as an environment variable instead.

RubeesEat/appsettings.json
```diff
@@ -1,9 +1,12 @@
...
-}
+{
+  "Logging": {
+    "LogLevel": {
+      "Default": "Information",
+      "Microsoft.AspNetCore": "Warning"
+    }
+  },
+  "ConnectionStrings": {
+    "DefaultConnectionString": "Server=localhost;Database=RubeesEatDB;UID=root;PW=;"
```
a-ctor, line 9: We don't use this configuration property anymore? We might want to consider using this configuration property instead of manually doing an `environement.getEnvironementVariable` as IConfiguration already supports overriding values using environment variables.

RubeesEat/Model/DB/DbTransactionExtensions.cs
```diff
@@ -0,0 +1,21 @@
...
+
+public static class DbTransactionExtensions
+{
+    public static DbCommand? CreateCommand(this DbTransaction transaction, string sql)
+    {
+        //Rider says transaction.Connection could be null
+        //(because we assign the connection to the transaction outside of this function)
+        //should we check for null and return null in that case?
+        var command = transaction.Connection?.CreateCommand();
+        if (command == null) return null;
```
a-ctor, line 14: Please no nullable return of `DbCommand`. This would imply that returning `null` is supported and we would need to check this fact on all usage sites. We can do a check and fail here for the connection to improve the error if the connection is not set, or we just assume that it is set and use `transaction.Connection!.CreateCommand()` instead

RubeesEat/Model/DB/DbCommandExtensions.cs
```diff
@@ -0,0 +1,43 @@
...
+    }
+
+    public static DbParameter AddParameter(this DbCommand command, string paramName, string paramValue)
+    {
+        var nameParam = command.CreateParameter();
+        nameParam.ParameterName = paramName;
+        nameParam.Value = paramValue;
+        command.Parameters.Add(nameParam);
+        return nameParam;
+    }
```
a-ctor, line 42: All of these methods do the same thing with the difference being the type of the value. `parameter.Value` is typed as `object` so we could do just one method with an `object value` and remove the remaining three methods.

By specifying the types here as separate values, we have the benefit of allowing only types that we know the parameters work for. So it is not a bad idea to do this to guard against supplying types of values that are not supported at all. If we want to keep the methods for this reason though, we should extract all the duplication into its own private method and call it from the four methods.

RubeesEat.IntegrationTests/DatabaseIntegrationTestBase.cs
```diff
@@ -0,0 +1,46 @@
...
+    private readonly string _connectionStringName = "DefaultConnectionString";
+
+    [SetUp]
+    public void SetUp()
+    {
+        string filePath = "./CreateTestDb.sql";
+        string sqlContent = File.ReadAllText(filePath);
+        string[] sqlCommands = sqlContent.Split([';'], StringSplitOptions.RemoveEmptyEntries);
+
+        // Trimmt leere Zeilen oder unnötige Leerzeichen von jedem Befehl
```
a-ctor, line 20: We can remove the obvious comment. There are cases where a comment is necessary to explain what is happening but a loop with 1 statement should be obvious enough for the comment to be unnecessary. Regarding comments, they are usually more necessary to explain why something is happening. If they explain what is happening, they are in danger of becoming out of date in the future when the next person forgets to update them.

a-ctor (changes requested):

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -0,0 +1,67 @@
...
+        using var command = connection.CreateCommand("INSERT INTO Persons (PersonID, FirstName, LastName) VALUES (@PersonID, @FirstName, @LastName);");
+        command.AddParameter("@PersonID", person.Id);
+        command.AddParameter("@FirstName", person.FirstName);
+        command.AddParameter("@LastName", person.LastName);
+        
+        connection.Open();
+        command.ExecuteNonQuery();
+    }
+
+
```
IremA99, line 40:
```suggestion
```

RubeesEat/Model/EqualityComparer/BillPropertyEqualityComparer.cs
```diff
@@ -0,0 +1,34 @@
...
+            return true;
+        if (x == null || y == null)
+            return false;
+        return x.Id == y.Id &&
+               x.Date == y.Date &&
+               x.Description == y.Description &&
+               x.EntryLines.SequenceEqual(y.EntryLines, _entryLineComparer);
+    }
+
+    public int GetHashCode([DisallowNull] Bill bill)
```
IremA99, line 20: Seems like the DisallowNull doesn't do anything here

RubeesEat/Model/EqualityComparer/EntryLinePropertyEqualityComparer.cs
```diff
@@ -0,0 +1,27 @@
...
+    {
+        if (ReferenceEquals(x, y))
+            return true;
+        if (x == null || y == null)
+            return false;
+        return _personPropertyComparer.Equals(x.Person, y.Person) &&
+               x.Amount == y.Amount;
+    }
+
+    public int GetHashCode([DisallowNull] EntryLine entryLine)
```
IremA99, line 18: Same as above, DisallowNull doesn't do anything

RubeesEat/Model/EqualityComparer/PersonPropertyEqualityComparer.cs
```diff
@@ -0,0 +1,28 @@
...
+        if (ReferenceEquals(x, y))
+            return true;
+        if (x == null || y == null)
+            return false;
+        return x.Id == y.Id &&
+               x.FirstName == y.FirstName &&
+               x.LastName == y.LastName;
+    }
+
+    public int GetHashCode([DisallowNull] Person person)
```
IremA99, line 18: Same as above

IremA99 (approved):

RubeesEat/Model/DB/DbCommandExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+
+    private static DbParameter AddParameterHelper(this DbCommand command, string name, object value)
+    {
+        var parameter = command.CreateParameter();
+        parameter.ParameterName = name;
+        parameter.Value = value;
+        command.Parameters.Add(parameter);
+        return parameter;
+    }
```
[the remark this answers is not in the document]
FelixSchoberRubi: According to Patricks review, the benefit here is that only certain Datatypes are allowed for the parameter "value", (in our case only Guid, decimal, DateTime and string are allowed). This provides extra security. If you want, we can remove it, so we lose security but get a simpler code.

RubeesEat/Model/EqualityComparer/BillPropertyEqualityComparer.cs
FelixSchoberRubi: It doesn't here. It's for static code analysis tools to know that no null should be entered. It tells the compiler that, so it can be useful if we ever accidentally enter a null value in our code

RubeesEat/Model/DB/DbCommandExtensions.cs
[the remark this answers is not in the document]
FelixSchoberRubi: Oh, but in this case the parameter never gets returned to the original caller, and we wanted to have them return the parameter, in case we need it. All the public AddParameter functions are void. The parameter that gets returned by the private helper function disappears because it is not used in a variable or returned further

RubeesEat.IntegrationTests/DB/DbBillRepositoryTest.cs
```diff
@@ -0,0 +1,95 @@
...
+    {
+        var persons = _dbPersonRepository.GetAll();
+        var bill = new Bill(Guid.NewGuid(), new DateTime(2025, 9, 9), "Mittagessen auf Item sein Nacken",
+        [
+            new EntryLine(persons[1], 1_000_000.50m),
+            new EntryLine(persons[2], -500_000.50m),
+            new EntryLine(persons[3], -500_000m)
+        ]);
+        _dbBillRepository.Add(bill);
+        Assert.That(_billPropertyComparer.Equals(_dbBillRepository.GetAll()[0], bill), Is.EqualTo(true));
```
a-ctor, line 51: Please fix other usages as well

```suggestion
        Assert.That(_dbBillRepository.GetAll()[0], Is.EqualTo(bill).Using(_billPropertyComparer));
```

a-ctor (approved):

PR #33, opened 2024-09-25
DSEAT-40 Rewrite tests using TagHelpers

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-40


KuJo7 (approved):

RubeesEat.IntegrationTests/PageObjects/UserHomepagePageObject.cs
```diff
@@ -14,13 +14,13 @@ public string[] GetBalanceChangeTexts()
 
     public BalanceChangeControlObject[] GetBalanceChanges()
     {
-        return FindElements(By.CssSelector(".entryLines > li"))
+        return FindElements("entryLines > li")
```
a-ctor, line 17: This does not look right. Why is there a CSS selector?

RubeesEat/Pages/BillDetails.cshtml
```diff
@@ -15,19 +15,19 @@
...
         <div class="bill-summary">
-            <div class="description">@item.Description</div>
-            <div class="personWhoPayed">Paid by: @(item.EntryLines[0].Person.FirstName)
+            <div class="description" t-id="description">@item.Description</div>
+            <div class="personWhoPaid" t-id="personWhoPaid">Paid by: @(item.EntryLines[0].Person.FirstName)
                 @(item.EntryLines[0].Person.LastName) </div>
-            <div class="date">@item.Date.ToString("D")</div>
-            <div class="amount">Total amount: @item.EntryLines[0].Amount€</div>
+            <div class="date" t-id="date">@item.Date.ToString("D")</div>
+            <div class="amount" t-id="amount">Total amount: @item.EntryLines[0].Amount€</div>
```
a-ctor, line 22: Ok before we forget removing the class names later on: Please remove all the other classes here and remove the CSS rule. We only set the margin so it might look bad for now but we will fix this with bootstrap later. This way, we won't forget removing these rules later on.

RubeesEat/Pages/BillDetails.cshtml
```diff
@@ -15,19 +15,19 @@
...
             @for (int i = 1; i < item.EntryLines.Length; i++)
             {
-                <div class="entryLine">
-                    <div class="person">@(item.EntryLines[i].Person.FirstName)
+                <div t-id="entryLine">
+                    <div class="person" t-id="person">@(item.EntryLines[i].Person.FirstName)
                         @(item.EntryLines[i].Person.LastName): </div>
-                    <div class="amountPerPerson">@(-item.EntryLines[i].Amount)€ </div>
+                    <div class="amountPerPerson" t-id="amountPerPerson">@(-item.EntryLines[i].Amount)€ </div>
                 </div>
```
a-ctor, line 31: Same here. Remove classes since we only use it for minor styling.

RubeesEat/Pages/Index.cshtml
```diff
@@ -12,21 +12,21 @@
...
                         @item.Amount.ToString("C", CultureInfo.CreateSpecificCulture("de-DE"))
                     </span>
-                    <span class="description">
+                    <span class="description" t-id="description">
                         @item.Description
                     </span>
-                    <span class="date">
+                    <span class="date" t-id="date">
                         @item.Date.ToString("D", CultureInfo.CreateSpecificCulture("en-US"))
                     </span>
```
a-ctor, line 31: Classes `description` and `date` are defined in `BillDetails.css` and should therefore add no styling here -> remove?

RubeesEat/Pages/ShowBills.cshtml
```diff
@@ -29,11 +29,11 @@
...
                     </span>
-                    <span class="description">
+                    <span class="description" t-id="description">
                         @item.Description
                     </span>
                 </div>
-                <div class="date">
+                <div class="date" t-id="date">
                     @item.Date.ToString("D", CultureInfo.CreateSpecificCulture("en-US"))
                 </div>
```
a-ctor, line 38: Same here. `description` and `date` should not have any styles because we define the styles in `BillDetails.css`. If that is true we can remove them here.

a-ctor (approved):

RubeesEat/Pages/Index.cshtml
IremA99: They are also defined in site.css and therefore do add styling

RubeesEat.IntegrationTests/PageObjects/UserHomepagePageObject.cs
IremA99: What do you mean exactly? We need the selector to get the entryLines displayed as a list.

a-ctor: I don't get why we can call `FindElements` with a css selector. In the current task we are supposed to clean up finding element by id so this change does not seem relevant.

RubeesEat/Pages/Index.cshtml
a-ctor: Okay then I leave it up to you. we can either break the styling now or remember to fix this when we switch to bootstrap.

RubeesEat/Pages/ShowBills.cshtml
a-ctor: Okay then I leave it up to you. we can either break the styling now or remember to fix this when we switch to bootstrap.

RubeesEat.IntegrationTests/PageObjects/UserHomepagePageObject.cs
FelixSchoberRubi: The line is a bit confusing: FindElements looks for a data-test-id, but in this case I set the data-test-id to "entryLines > li" (so that it is clear from the code, that it is looking for lis in entryline). I can change the data-test-id to something like "entrylinesLine" or something like that, if that is better.

But since you said we only want to clean up finding by id, we can revert this line to look for CssSelector, as before

a-ctor: It should not look like a CSS selector if it is not a CSS selector. We can use a singular name `entryLine` or `balanceChange` as ID for each element and search for that instead.

a-ctor (approved):

PR #38, opened 2024-10-07
Feature/dseat 44 add is active to person

IremA99:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-44


RubeesEat/Model/EntryLine.cs
```diff
@@ -8,6 +8,11 @@ public class EntryLine
     public EntryLine(Person person, decimal amount)
     {
         ArgumentNullException.ThrowIfNull(person);
+
+        if (!person.IsActive)
+        {
+            throw new ArgumentException("Person is not active!");
+        }
```
FelixSchoberRubi, line 15: Why throw Exception? We have bills where the entryline is from an inactive user, e.g. if you show all bills, there will be many old bills with inactive users. If we throw an exception, the website will just stop working.

FelixSchoberRubi (changes requested):
What is missing, which was described in the ticket, is only showing active users in the dropbox for splitBill. Also we need tests for isActive

FelixSchoberRubi (approved):
Looks good!
Since the excel branch is already merged, we need to rebase on main and then correctly insert the persons with their isActive status, according to the PersonDTO

PR #41, opened 2024-10-09
Feature/dseat 22 open id auth

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-22

### DIFF
--- a/RubeesEat/Controllers/BillController.cs
+++ b/RubeesEat/Controllers/BillController.cs
@@ -1,4 +1,5 @@
 using System.Globalization;
+using Microsoft.AspNetCore.Authorization;
 using Microsoft.AspNetCore.Http.HttpResults;
 using Microsoft.AspNetCore.Mvc;
 using RubeesEat.Model;
--- a/RubeesEat/Startup.cs
+++ b/RubeesEat/Startup.cs
@@ -26,9 +26,9 @@ public class Startup
                 .AddOpenIdConnect(options =>
                 {
                     options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
-                    options.ClientId = "BdOG7g762eoAF7sXkoH2Gu2UXBVwzViJ";
-                    options.ClientSecret = "_pIXrOgfWvvFhOqUvuBMSASLhL2RpryArgEGdDRfWWCxnFVp7aqEGRt7_8lcBHkF";
-                    options.Authority = "https://dev-amcvh6f04mjr767m.us.auth0.com";
+                    options.ClientId = Configuration["OpenId:ClientId"];
+                    options.ClientSecret = Configuration["OpenId:ClientSecret"];
+                    options.Authority = Configuration["OpenId:Authority"];
                     options.GetClaimsFromUserInfoEndpoint = true;
                 });
         services.AddAuthorization(option =>
--- a/RubeesEat/Startup.cs
+++ b/RubeesEat/Startup.cs
@@ -1,4 +1,6 @@
 ﻿using System.Collections.Immutable;
+using Microsoft.AspNetCore.Authentication.Cookies;
+using Microsoft.AspNetCore.Authentication.OpenIdConnect;
 using RubeesEat.Model;
 using RubeesEat.Model.DB;
 
@@ -15,6 +17,24 @@ public class Startup
 
     public void ConfigureServices(IServiceCollection services)
     {
+        services.AddAuthentication(options =>
+                {
+                    options.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                    options.DefaultChallengeScheme = OpenIdConnectDefaults.AuthenticationScheme;
+                })
+                .AddCookie()
+                .AddOpenIdConnect(options =>
+                {
+                    options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                    options.ClientId = "BdOG7g762eoAF7sXkoH2Gu2UXBVwzViJ";
+                    options.ClientSecret = "_pIXrOgfWvvFhOqUvuBMSASLhL2RpryArgEGdDRfWWCxnFVp7aqEGRt7_8lcBHkF";
+                    options.Authority = "https://dev-amcvh6f04mjr767m.us.auth0.com";
+                    options.GetClaimsFromUserInfoEndpoint = true;
+                });
+        services.AddAuthorization(option =>
+        {
+            option.FallbackPolicy = option.DefaultPolicy;
+        });
         services.AddRazorPages();
 
         var connectionString = Configuration["ConnectionStrings:DefaultConnectionString"];
@@ -41,6 +61,8 @@ public class Startup
         app.UseStaticFiles();
 
         app.UseRouting();
+        app.UseAuthentication();
+        app.UseAuthorization();
 
         app.UseEndpoints(
             endpoints =>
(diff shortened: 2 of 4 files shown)


RubeesEat/Startup.cs
```diff
@@ -1,20 +1,48 @@
...
+                        options.DefaultAuthenticateScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                        options.DefaultChallengeScheme = OpenIdConnectDefaults.AuthenticationScheme;
+                    })
+                    .AddCookie()
+                    .AddOpenIdConnect(options =>
+                    {
+                        options.SignInScheme = CookieAuthenticationDefaults.AuthenticationScheme;
+                        options.ClientId = Configuration["OpenId:ClientId"];
+                        options.ClientSecret = Configuration["OpenId:ClientSecret"];
+                        options.Authority = Configuration["OpenId:Authority"];
```
a-ctor, line 36: Please move all these settings into a class `OpenIDSettings` that we read from the configuration. This way, we can add validation attributes to ensure that all the fields are set correctly. Right now, we hope that the settings are correct.

RubeesEat/Startup.cs
```diff
@@ -1,20 +1,48 @@
...
-    public Startup(IConfiguration configuration)
+    public Startup(IConfiguration configuration, IHostingEnvironment environment)
     {
         Configuration = configuration;
+        _environment = environment;
     }
 
     public void ConfigureServices(IServiceCollection services)
     {
+        if (!_environment.IsEnvironment("Test"))
```
a-ctor, line 23: Please don't. We do not want to add code in here that is just for the tests. Ideally, we can switch out the authentication in the tests by replacing the services when we start the server in the test (see https://learn.microsoft.com/en-us/aspnet/core/test/integration-tests?view=aspnetcore-8.0#mock-authentication). Please try to rewrite this and write it so that the test can switch out the user it wants to use for requests.

RubeesEat/Controllers/BillController.cs
```diff
@@ -1,4 +1,5 @@
 using System.Globalization;
+using Microsoft.AspNetCore.Authorization;
```
a-ctor, line 2: revert?

RubeesEat.IntegrationTests/IntegrationTestBase.cs
```diff
@@ -18,7 +18,8 @@ public void SetUp()
     {
         TestServer = new TestServer(new WebHostBuilder()
                                     .UseStartup<Startup>()
-                                    .ConfigureTestServices(ConfigureServices));
+                                    .ConfigureTestServices(ConfigureServices)
+                                    .UseEnvironment("Test"));
```
a-ctor, line 22: We should be able to revert this and do the things we need when we configure the test services.

a-ctor (dismissed):

RubeesEat/appsettings.json
```diff
@@ -10,5 +10,10 @@
     "DefaultConnectionString": "Server=localhost;Database=RubeesEatDB;UID=root;PWD=;",
     "TestConnectionString": "Server=localhost;Database=RubeesEatTestDB;UID=root;PWD=;"
   },
+  "OpenId": {
+    "ClientId": "%OpenId__ClientId%",
+    "ClientSecret": "%OpenId__ClientSecret%",
+    "Authority": "%OpenId__Authority%"
```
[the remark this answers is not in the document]
FelixSchoberRubi: I"m assuming because we had placeholders/default values for the connectionsstrings (which are anyhow replaced by the env variables).

Should we just remove these alltogether? It works without the OpenID part in the appsettings, since it is anyhow gotten from the env variables.

RubeesEat/Startup.cs
FelixSchoberRubi: I could put the Required attribute for validation on the fields of the class, if thats what you mean. I don't know what other validation attributes i can set in this case.

IremA99 (approved):

KuJo7 (approved):

RubeesEat.IntegrationTests/SeleniumTests/OpenIDSeleniumTests.cs
```diff
@@ -0,0 +1,15 @@
+using RubeesEat.IntegrationTests.PageObjects;
+
+namespace RubeesEat.IntegrationTests.SeleniumTests;
+
+[TestFixture]
+[Explicit("does not run on teamcity")]
```
MichaelKetting, line 6: Why does it not run on teamcity? Is it because we can't reach that external service listed in line 13? Who is responsible for maintaining this service?

I think there's  OpenID stubs for local testing, should ask the LF Team. @a-ctor we should look into this.
cc @Frazerus

MichaelKetting (approved):
I have reviewed this based on the comments made by @a-ctor but I'm not sure if I considered all cavears / wishes / intents since I don't know which files are new / already reviewed by @a-ctor. /crossing fingers I didn't miss anything relevant that should ahve been done differently/

RubeesEat.IntegrationTests/OpenIDIntegrationTestBase.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Microsoft.Extensions.Hosting;
+using RubeesEat.IntegrationTests.WebTesting;
+using RubeesEat.Model;
+
+namespace RubeesEat.IntegrationTests;
+
+public class OpenIDIntegrationTestBase
```
MichaelKetting, line 12: This type should be SeleniumIntegrationTestBase.
The SeleniumIntegrationTestBase should derive and replace / override relevant features and be named in a sensible way.

MichaelKetting (changes requested):

RubeesEat.IntegrationTests/SeleniumTests/OpenIDSeleniumTests.cs
FelixSchoberRubi: I set the env variables in teamcity, and the test ran successfully. Until we get the stubs we can do it this way, with auth0

RubeesEat.IntegrationTests/TestBases/IntegrationTestBase.cs
FelixSchoberRubi, line 1: The only change i did here was move this file into the new TestBases folder

RubeesEat.IntegrationTests/OpenIDIntegrationTestBase.cs
FelixSchoberRubi: Now i have the following inheritance structure:
SeleniumIntegrationTestBase > WithoutAuth > WithTestPerson > WithTestBills

I'm writing this here to make it easier to review since Github does not display the files in the order that makes sense for the inheritance.

RubeesEat.IntegrationTests/TestBases/Selenium/WithTestBills.cs
```diff
@@ -0,0 +1,52 @@
+using System.Collections.Immutable;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using RubeesEat.Model;
+
+namespace RubeesEat.IntegrationTests;
+
+public class WithTestBills : WithTestPersons
```
MichaelKetting, line 8: Interesting test base name. I recommend making them abstract. As to naming, I can see it. I hope @a-ctor likes it, too.

MichaelKetting (approved):
Thanks for the comments. I actually managed to reconstruct in memory before seeing the comments :)

PR #42, opened 2024-10-09
Feature/dseat 46 assign users to their accounts

FelixSchoberRubi:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-46

### DIFF
--- a/ExcelToDBTransferer/PersonDTO.cs
+++ b/ExcelToDBTransferer/PersonDTO.cs
@@ -9,18 +9,20 @@ public class PersonDTO
     public string LastName { get; }
     public string Initials { get;}
     public bool IsActive { get; }
+    public string? LoginName { get; }
 
-    public PersonDTO(string firstName, string lastName, string initials, bool isActive)
+    public PersonDTO(string firstName, string lastName, string initials, bool isActive, string? loginName)
     {
         Id = Guid.NewGuid();
         FirstName = firstName;
         LastName = lastName;
         Initials = initials;
         IsActive = isActive;
+        LoginName = loginName;
     }
     
     public Person MapToPerson()
     {
-        return new Person(Id, FirstName, LastName);
+        return new Person(Id, FirstName, LastName, LoginName, IsActive);
     }
 }
--- a/RubeesEat.IntegrationTests/CreateTestDb.sql
+++ b/RubeesEat.IntegrationTests/CreateTestDb.sql
@@ -7,7 +7,8 @@ CREATE TABLE Persons
     PersonID VARCHAR(64) PRIMARY KEY,
     FirstName VARCHAR(64) NOT NULL,
     LastName VARCHAR(64) NOT NULL,
-    IsActive BOOLEAN NOT NULL DEFAULT TRUE
+    IsActive BOOLEAN NOT NULL DEFAULT TRUE,
+    LoginName VARCHAR(64)
 );
 
 CREATE TABLE Bills
@@ -26,16 +27,14 @@ CREATE TABLE EntryLines
     FOREIGN KEY (BillID) REFERENCES Bills(BillID)
 );
 
-INSERT INTO Persons (PersonID, FirstName, LastName)
-VALUES ('A05764E0-C2F5-4A3F-8F04-746AEE8B355B', 'Item', 'Arslan');
-INSERT INTO Persons (PersonID, FirstName, LastName)
-VALUES ('883703F3-EEA8-4BCE-BACD-4A77FFE0C294', 'Patrick', 'Widener');
-INSERT INTO Persons (PersonID, FirstName, LastName)
-VALUES ('544FEBD0-05F8-471A-BAFC-CA7135538031', 'Lilli', 'Grubber');
-INSERT INTO Persons (PersonID, FirstName, LastName)
-VALUES ('7DBF157B-CBFF-43CE-BFDD-B367611BB1A5', 'Mich', 'Ludwig');
-INSERT INTO Persons (PersonID, FirstName, LastName)
-VALUES ('8e3edf3b-7dcd-4815-bf16-4bc9b9d6dd3b', 'No', 'Bills');
+INSERT INTO Persons (PersonID, FirstName, LastName, LoginName)
+VALUES ('A05764E0-C2F5-4A3F-8F04-746AEE8B355B', 'Item', 'Arslan', 'item.arslan@test.com');
+INSERT INTO Persons (PersonID, FirstName, LastName, LoginName)
+VALUES ('883703F3-EEA8-4BCE-BACD-4A77FFE0C294', 'Patrick', 'Widener', 'patrick.widener@test.com');
+INSERT INTO Persons (PersonID, FirstName, LastName, LoginName)
+VALUES ('544FEBD0-05F8-471A-BAFC-CA7135538031', 'Lilli', 'Grubber', 'lilli.grubber@test.com');
+INSERT INTO Persons (PersonID, FirstName, LastName, LoginName)
+VALUES ('7DBF157B-CBFF-43CE-BFDD-B367611BB1A5', 'Mich', 'Ludwig', 'mich.ludwig@test.com');
 
 INSERT INTO Bills(BillID, Description, Date)
 VALUES ('91CD57E6-418E-4628-82CC-09D471153CF6', 'Mittagessen auf Patrick sein Nacken', '2024-09-09 10:34:09');
--- a/RubeesEat/Model/DB/DbBillRepository.cs
+++ b/RubeesEat/Model/DB/DbBillRepository.cs
@@ -76,7 +76,8 @@ public class DbBillRepository(IDbConnectionFactory connectionFactory) : IBillRep
                                                       	E.Amount,
                                                       	P.PersonID,
                                                       	P.FirstName,
-                                                      	P.LastName
+                                                      	P.LastName,
+                                                      	P.LoginName
                                                       FROM
                                                       	Bills B
                                                       INNER JOIN 
@@ -186,7 +187,8 @@ public class DbBillRepository(IDbConnectionFactory connectionFactory) : IBillRep
                                        E.Amount,
                                        P.PersonID,
                                        P.FirstName,
-                                       P.LastName
+                                       P.LastName,
+                                       P.LoginName
                                    FROM
                                        Bills B
                                    INNER JOIN
@@ -223,7 +225,8 @@ public class DbBillRepository(IDbConnectionFactory connectionFactory) : IBillRep
             var person = new Person(
                 reader.GetGuid(5),
                 reader.GetString(6),
-                reader.GetString(7)
+                reader.GetString(7),
+                reader.IsDBNull(8) ? null : reader.GetString(8)
             );
 
             var entryLine = new EntryLine(
--- a/RubeesEat/Model/DB/DbPersonRepository.cs
+++ b/RubeesEat/Model/DB/DbPersonRepository.cs
@@ -1,11 +1,21 @@
+using System.Data;
+using System.Security.Claims;
+
 namespace RubeesEat.Model.DB;
 
-public class DbPersonRepository(IDbConnectionFactory connectionFactory) : IPersonRepository
+public class DbPersonRepository: IPersonRepository
 {
+    private readonly IDbConnectionFactory _connectionFactory;
+    
+    public DbPersonRepository(IDbConnectionFactory connectionFactory)
+    {
+        _connectionFactory = connectionFactory;
+    }
+    
     public IReadOnlyList<Person> GetAll()
     {
-        using var connection = connectionFactory.CreateDbConnection();
-        using var command = connection.CreateCommand("SELECT PersonID, FirstName, LastName, IsActive FROM Persons ORDER BY LastName ASC;");
+        using var connection = _connectionFactory.CreateDbConnection();
+        using var command = connection.CreateCommand("SELECT PersonID, FirstName, LastName, LoginName, IsActive FROM Persons ORDER BY LastName ASC;");
         connection.Open();
 
         using var reader = command.ExecuteReader();
@@ -18,7 +28,9 @@ public class DbPersonRepository(IDbConnectionFactory connectionFactory) : IPerso
             (
                 reader.GetGuid(0),
                 reader.GetString(1),
-                reader.GetString(2)
+                reader.GetString(2),
+                reader.IsDBNull(3) ? null : reader.GetString(3),
+                reader.GetBoolean(4)
             ));
         }
 
@@ -27,8 +39,8 @@ public class DbPersonRepository(IDbConnectionFactory connectionFactory) : IPerso
 
     public IReadOnlyList<Person> GetAllActive()
     {
-        using var connection = connectionFactory.CreateDbConnection();
-        using var command = connection.CreateCommand("SELECT PersonID, FirstName, LastName, IsActive FROM Persons WHERE IsActive = 1 ORDER BY LastName ASC;");
+        using var connection = _connectionFactory.CreateDbConnection();
+        using var command = connection.CreateCommand("SELECT PersonID, FirstName, LastName, LoginName, IsActive FROM Persons WHERE IsActive = 1 ORDER BY LastName ASC;");
         connection.Open();
 
         using var reader = command.ExecuteReader();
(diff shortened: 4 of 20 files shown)


ExcelToDBTransferer/PersonDTO.cs
```diff
@@ -9,14 +9,16 @@ public class PersonDTO
     public string LastName { get; }
     public string Initials { get;}
     public bool IsActive { get; }
+    public string LoginName { get; }
 
-    public PersonDTO(string firstName, string lastName, string initials, bool isActive)
+    public PersonDTO(string firstName, string lastName, string initials, bool isActive, string loginName)
     {
```
a-ctor, line 15: Shouldn't this be nullable as we wanted inactive users to not have a login name?

RubeesEat/Model/DB/DbBillRepository.cs
```diff
@@ -162,15 +162,14 @@ public decimal GetBalance(Person user)
...
+        var result = command.ExecuteScalar();
+        
+        if (result == DBNull.Value)
         {
-            sum += reader.GetDecimal(0);
+            return 0m;
         }
 
-        return sum;
+        return Convert.ToDecimal(result);
```
a-ctor, line 172: Why do we need this change?

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -27,24 +40,53 @@ public IReadOnlyList<Person> GetAll()
...
+            return new Person
+            (
+                reader.GetGuid(0),
+                reader.GetString(1),
+                reader.GetString(2)
+            );
+        }
+        var newPerson = new Person(Guid.NewGuid(), firstName, lastName, loginName);
+        Add(newPerson);
+        return newPerson;
```
a-ctor, line 84: Please extract this into its own method `Person? GetByLoginName(string loginName)` we can then share the mapping logic with `GetById`

RubeesEat/Model/Person.cs
```diff
@@ -8,8 +8,9 @@ public class Person
     public Guid Id { get; }
     public string FirstName { get; }
     public string LastName { get; }
+    public string LoginName { get; }
```
a-ctor, line 11: This should be nullable and you have a warning at the assignment

RubeesEat/Model/Person.cs
```diff
@@ -8,8 +8,9 @@ public class Person
...
 
-    public Person(Guid id, string firstName, string lastName)
+    public Person(Guid id, string firstName, string lastName, string loginName = null)
```
a-ctor, line 13: We are also missing nullability. Please don't assign a default value as the default is not "User with no login name". Instead, users should have a login name and some don't because they are no longer active or whatever.

We also have the `Create` method above that does not have the added login name parameter.

a-ctor (changes requested):
We are also missing the SQL scripts and the update for the login name. Please also ensure that the login name has a unique constraint.

RubeesEat/Model/DB/DbBillRepository.cs
FelixSchoberRubi: This is a bug that has been here for a while but wasn't noticed. When the current user has no bills, and the website wants to display the balance, the sql statement SUM then returns Null, instead of 0, and the function breaks when trying to convert to decimal

When a users login name is not in the db, a new Person is created for them, with no bills yet, and this bug shows up. This is why it was noticed in this branch.

So i fixed the bug here in a seperate commit. Now the function returns 0, when the sql query returns null.

RubeesEat/Model/Person.cs
FelixSchoberRubi: Since we are changing the function signature, we will have to change it for each call of the constructor or Create function (alltogether 63 usages, especially in tests). Is there a way to solve that other than manually adjusting each usage?

FelixSchoberRubi: Since we are changing the function signature, we will have to change it for each call of the constructor or Create function (alltogether 63 usages, especially in tests). Is there a way to solve that other than manually adjusting each usage?

FelixSchoberRubi: Never mind, I took care of it

IremA99 (approved):

RubeesEat/Model/Person.cs
```diff
@@ -4,13 +4,14 @@ namespace RubeesEat.Model;
 
 public class Person
 {
-    public static Person Create(string firstName, string lastName) => new(Guid.NewGuid(), firstName, lastName, true);
+    public static Person Create(string firstName, string lastName, string? loginName) => new(Guid.NewGuid(), firstName, lastName, loginName);
     public Guid Id { get; }
     public string FirstName { get; }
     public string LastName { get; }
     public bool IsActive { get; }
+    public string? LoginName { get; }
```
MichaelKetting, line 12:
```suggestion
    [MemberNotNullWhen(true, nameof(LoginName ))]
    public bool IsActive { get; }
    public string? LoginName { get; }
```

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -1,11 +1,25 @@
...
+    {
+        _httpContextAccessor = httpContextAccessor;
+        _connectionFactory = connectionFactory;
+    }
+    public DbPersonRepository(IDbConnectionFactory connectionFactory)
+    {
+        _connectionFactory = connectionFactory;
+    }
     public IReadOnlyList<Person> GetAll()
     {
```
MichaelKetting, line 20: There should be empty lines between the method declarations and fields/method declaration.

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -18,17 +32,20 @@ public IReadOnlyList<Person> GetAll()
...
+                reader.IsDBNull(3) ? null : reader.GetString(3),
+                reader.GetBoolean(4)
+
             ));
         }
 
         return persons.AsReadOnly();
     }
-
+
```
MichaelKetting, line 44: revert whitespace characters.

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -18,17 +32,20 @@ public IReadOnlyList<Person> GetAll()
             (
                 reader.GetGuid(0),
                 reader.GetString(1),
-                reader.GetString(2)
+                reader.GetString(2),
+                reader.IsDBNull(3) ? null : reader.GetString(3),
+                reader.GetBoolean(4)
+
```
MichaelKetting, line 38: remove blank line at end of loop block.

RubeesEat/Model/DB/DbBillRepository.cs
```diff
@@ -224,7 +225,8 @@ private IReadOnlyList<Bill> CreateBillsFromReader(DbDataReader reader)
             var person = new Person(
                 reader.GetGuid(5),
                 reader.GetString(6),
-                reader.GetString(7)
+                reader.GetString(7),
+                reader.IsDBNull(8) ? null : reader.GetString(8)
             );
```
MichaelKetting, line 230: Not somehting we will change now and I don't know what input @a-ctor had about it, but generally speaking, having the SELECT projection and the column access separated between methods isn't good since it means the information about the column order is lost in the call chain. It is possible that there's no good alternative without using an O/R-mapper.

MichaelKetting (changes requested):
Please let me know if something was already discussed and decided differently.

RubeesEat/Model/DB/DbBillRepository.cs
FelixSchoberRubi: Irem and Joel rebased on main, as preparation for merging later. Unfortunately I forgot about this commit, so I couldn't tell them to leave this commit alone.

RubeesEat/Model/Person.cs
```diff
@@ -4,13 +4,14 @@ namespace RubeesEat.Model;
 
 public class Person
 {
-    public static Person Create(string firstName, string lastName) => new(Guid.NewGuid(), firstName, lastName, true);
+    public static Person Create(string firstName, string lastName, string? loginName) => new(Guid.NewGuid(), firstName, lastName, loginName);
```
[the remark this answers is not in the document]
FelixSchoberRubi: We need to be able to create inactive users as well, to display old bills with entrylines of users that are no longer active

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -65,28 +84,65 @@ public void Add(Person person)
     
     public Person GetCurrentUser()
     {
-        return GetAll()[^1];
+        var httpContext = _httpContextAccessor.HttpContext;
+        if (httpContext == null)
+            throw new NullReferenceException("HttpContext is null");
+        var loginName = httpContext.User.FindFirst("name")?.Value;
+        var nickName = httpContext.User.FindFirst("nickname")?.Value;
```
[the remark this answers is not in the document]
FelixSchoberRubi: Yeah, the only reason I did it in this way, was to avoid having to change the function signature. But I can change it so that it passes the IPrinicipal instead. I think this is the only place where we use HttpContext in this way.

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -65,28 +84,65 @@ public void Add(Person person)
...
+        var splitNickName = nickName.Split(".");
+        var firstName = splitNickName[0];
+        var lastName = splitNickName[1];
+        if (loginName == null)
+            throw new NullReferenceException("User is inactive");
```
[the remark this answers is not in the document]
FelixSchoberRubi: Yes, Patrick pointed this out to us. NullReferenceExceptions shouldn't be thrown on purpose because the compiler throws them, so if you throw them on purpose it just makes understanding where the error message comes from more difficult.

RubeesEat/Model/DB/DbPersonRepository.cs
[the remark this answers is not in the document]
MichaelKetting: From an outside perspective, this sounds like a sensible change. I can't speak to the implication for getting this done in time.

RubeesEat/Model/DB/DbBillRepository.cs
FelixSchoberRubi: Is extracted: https://github.com/re-motion/students-RubeesEat/pull/44

RubeesEat/Model/DB/DbPersonRepository.cs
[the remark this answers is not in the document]
MichaelKetting: This isn't updated as of the time I got this in re-review so I'll keep the convo open until it's fixed :)

MichaelKetting (changes requested):

RubeesEat/Pages/Index.cshtml
```diff
@@ -3,10 +3,10 @@
 @inject IBillRepository billRepository
 @inject IPersonRepository personRepository
 
-<h1>Hello, @personRepository.GetCurrentUser().FirstName!</h1>
+<h1>Hello, @personRepository.GetOrCreateUser(HttpContext.User).FirstName!</h1>
```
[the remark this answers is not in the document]
FelixSchoberRubi: Funnily enough, I actually already wrote a ticket for this yesterday. I originally mentioned it in a ticket for editing bills because there we use CurrentUser even more, and I discussed with Irem that we should make it into a seperate ticket. https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-64

My original thought was to add it to the personRepo, so that when you use GetOrCreateUser, it first checks if it already queried it before. Should we do it in the controller instead?

RubeesEat/Model/InMemory/InMemoryPersonRepository.cs
```diff
@@ -37,7 +39,7 @@ public void Add(Person person)
         _persons.Add(person.Id, person);
     }
 
-    public Person GetCurrentUser()
+    public Person GetOrCreateUser(ClaimsPrincipal user)
     {
         return _persons.TryGetValue(CurrentUser, out var person) 
             ? person
```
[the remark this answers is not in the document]
FelixSchoberRubi: Originally InMemoryPersonRepository was the only implementation of the IPersonRepository Interface, until we implemented the database. So yeah, now it is just relegated to Testing.

That's why the function looks weird: The interface needed to have the ClaimsPrincipal as a parameter, but the InMemory implementation doesn't need it.

We can move the InMemoryPersonRepository to the Unit Tests Project in a seperate ticket/PR, to keep it seperate from this PR.

RubeesEat/Model/DB/DbPersonRepository.cs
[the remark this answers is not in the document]
FelixSchoberRubi: Sorry, I somehow overlooked it and thought it was showing a change that was already made, instead of the code suggestion from you

RubeesEat/Pages/Index.cshtml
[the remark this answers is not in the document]
MichaelKetting: Good catch on the ticket! Happy you thought of it first :)

Repo vs Controller: the user is something that's just for a specific request. the Repo lives based on what's configured in the IoC provider so if you cache it there you might end up with global state (TM). that leads to many tears for the developer as well as tears in the fabric of the space time continuum.

RubeesEat/Model/InMemory/InMemoryPersonRepository.cs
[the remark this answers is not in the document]
MichaelKetting: that's okay.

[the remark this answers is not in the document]
FelixSchoberRubi: new ticket was created https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-67

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -41,52 +53,88 @@ public IReadOnlyList<Person> GetAllActive()
             (
                 reader.GetGuid(0),
                 reader.GetString(1),
-                reader.GetString(2)
+                reader.GetString(2),
+                reader.IsDBNull(3) ? null : reader.GetString(3),
+                reader.GetBoolean(4)
+
             ));
```
MichaelKetting, line 60:
```suggestion
                reader.GetBoolean(4)
            ));
```

RubeesEat/Model/DB/DbPersonRepository.cs
```diff
@@ -41,52 +53,88 @@ public IReadOnlyList<Person> GetAllActive()
...
+    public Person GetOrCreateUser(ClaimsPrincipal user)
     {
-        return GetAll()[^1];
+        var loginName = user.FindFirst("name")?.Value;
+        var nickName = user.FindFirst("nickname")?.Value;
+        var splitNickName = nickName.Split(".");
+        var firstName = splitNickName[0];
+        var lastName = splitNickName[1];
+        if (loginName == null)
+            throw new InvalidOperationException("User is inactive");
```
MichaelKetting, line 88: I believe the exception message is not correct. you're talking about a ClaimsPrincipal here. That's an object you got from the authentication infrastructure so there's a user sitting in front of the system. And for some reason, this user did not contain a metadata `name` in its description/bag/whatever-you-call-it. The excpetion should tell you this.

MichaelKetting (changes requested):
Request since I want to se the message :)

Also, there should be a unit test for this.

RubeesEat/Model/DB/DbPersonRepository.cs
FelixSchoberRubi: This is one of those cases where we discussed something specific with Patrick. The idea is that if the loginName is null, this means (in our application) that the User is inactive. https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-46

If there is a user sitting in front of the system getting this error message, they should go to an admin and let them know that their user is inactive.

MichaelKetting: I believe there's a misunderstanding. 
Type `User` propety `LoginName` being `null` is exactly what's discussed with Patrick.  
Type `ClaimsPrincipal` not having a `name` entry in its proprties is in my understanding a completely different scenario because it means that the OpenID provider did not send all the fields we expect for the user.

FelixSchoberRubi: I see what you mean. 
I think in this case it's a little strange because we get the loginName from the name entry. We did not know how the loginName would be given in our specific case, when it comes to the company's OpenID provider. Our first guess was that the email address would be the loginName, but for some reason the ClaimsPrincipal did not automatically have the email entry and there was some complication there. So we talked to Patrick and we settled on using the name field for now. Patrick said it didn't matter for the time because we anyhow have to change this once we know exactly which field corresponds to the loginName.

But you mean it should always send us the entries we expect, even when the user is inactive, right? In that case it seems like we would never get the case that we can tell if an inactive user is logging in. If inactive users have no loginName in our db, and such an inactive user were to login, but have a loginName in their ClaimsPrincipal, then the program will just create a new user (since it cannot find the right entry in the db), which is not what we want.

MichaelKetting: > But you mean it should always send us the entries we expect, even when the user is inactive, right? 

More precisely: at this point, the concept of the user being inactive isn't relevant since it's not part of the decision whether to throw an exception. If you want to deny login to inactive users, you would need to check if the user exists and is inactive and if so, then throw a matching exception. For now, as the code stands, we truly only check whether we are getting valid data from the authentiction provider. different subject matter.

>  In that case it seems like we would never get the case that we can tell if an inactive user is logging in.
Correct, see above. might be a new ticket.

>  If inactive users have no loginName in our db, and such an inactive user were to login, but have a loginName in their ClaimsPrincipal, then the program will just create a new user (since it cannot find the right entry in the db), which is not what we want.
True. maybe the concept of having no login-name isn't helpful for inactive users? The problem only exists for legacy data. there, we could simply populate with firstname.lastname@rubicon.eu (with or with out the rubicon suffix) as the login name and then we always have helpful data. Also a separate task.

Either way, this exception right here only valdiates if we get all the data we get from the authentication provider is formatted correctly. if it's not, there is an exception because we can only reject the user. Likely this would be a total system  outtage because all users would be logging in using an invalid data format.

FelixSchoberRubi: I created a new ticket regarding the inactive User problem: https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-70

I will rename the exception properly, that the claim does not have the proper information

FelixSchoberRubi:
I added unit tests for GetOrCreateCurrentUser() (also deleted an old test for the older version of the function. The old function behaved very strangely, simply returning the first person after using GetAll, so the old test was unnecessary).

Added more exceptions, in case of other ways claims could be unexpected (missing certain fields or the field being in a different formating). This part may be changed once we are certain how a Claim and its fields will look like in our specific case.

RubeesEat.IntegrationTests/CreateTestDb.sql
```diff
@@ -7,7 +7,8 @@ CREATE TABLE Persons
     PersonID VARCHAR(64) PRIMARY KEY,
     FirstName VARCHAR(64) NOT NULL,
     LastName VARCHAR(64) NOT NULL,
-    IsActive BOOLEAN NOT NULL DEFAULT TRUE
+    IsActive BOOLEAN NOT NULL DEFAULT TRUE,
+    LoginName VARCHAR(64)
```
MichaelKetting, line 11: We won't change this now but there's a general best practice to not use power-of-two length values in data design. No human understands why a name may only have 63 or 64 characters. Every human understands why it's 50 or a 100 characters.

PS: yes, I'm aware about the meaning of 'human' ;)

MichaelKetting, line 11: Don't know if you already did: please add a ticket that Person.LoginName must not be null in database or domain. This is a consequence of the login/inactive-user topic.

MichaelKetting (approved):

PR #43, opened 2024-10-16
Feature/dseat 50 allow editing bills

IremA99:
https://re-motion.atlassian.net/jira/software/projects/DSEAT/boards/17?selectedIssue=DSEAT-50

### DIFF
--- a/RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
+++ b/RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
@@ -38,6 +38,45 @@ public class BillControllerTest : IntegrationTestBase
         Assert.That(billsAfter.Length, Is.EqualTo(bills.Length + 1));
         Assert.That(billsAfter[^1].Description, Is.EqualTo("test234"));
     }
+
+    [Test]
+    public async Task Update_DoesUpdate()
+    {
+        var resultPersons = await HttpClient.GetStringAsync("/api/persons");
+        var persons = JsonConvert.DeserializeObject<Person[]>(resultPersons)!;
+        var resultBills = await HttpClient.GetStringAsync("/api/bills");
+        var bills = JsonConvert.DeserializeObject<Bill[]>(resultBills)!;
+
+        var existingBill = bills[0];
+        var url = $"/api/bills/{existingBill.Id}";
+        var newDescription = "Updated Description";
+        
+        var content = new FormUrlEncodedContent(new[]
+        {
+            new KeyValuePair<string, string>("billId", existingBill.Id.ToString()),
+            new KeyValuePair<string, string>("billDescription", newDescription),
+            new KeyValuePair<string, string>("billAmount", "20.00"),
+            
+            new KeyValuePair<string, string>("id0", persons[0].Id.ToString()),
+            new KeyValuePair<string, string>("amount0", "10.00"), 
+            
+            new KeyValuePair<string, string>("id1", persons[1].Id.ToString()),
+            new KeyValuePair<string, string>("amount1", "10.00"),
+        });
+
+        var response = await HttpClient.PostAsync(url, content);
+
+        Assert.That(response.StatusCode, Is.EqualTo(HttpStatusCode.Found));
+
+        var resultBillsAfter = await HttpClient.GetStringAsync("/api/bills");
+        var billsAfter = JsonConvert.DeserializeObject<Bill[]>(resultBillsAfter)!;
+        var updatedBill = billsAfter.FirstOrDefault(b => b.Id == existingBill.Id);
+
+        Assert.That(updatedBill.Description, Is.EqualTo(newDescription));
+        Assert.That(updatedBill.EntryLines[0].Amount, Is.EqualTo(20.00));
+        Assert.That(updatedBill.EntryLines[1].Amount, Is.EqualTo(-10.00));
+    }
+
     
     [Test]
     public async Task? Create_DuplicatePerson_BadRequest()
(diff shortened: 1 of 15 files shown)


RubeesEat.IntegrationTests/PageObjects/BillDetailsPageObject.cs
```diff
@@ -20,4 +20,10 @@ public EntryLineControlObject[] GetEntryLines()
                .Select(e => new EntryLineControlObject(this, e))
                .ToArray();
     }
+    
+    public EditBillPageObject ClickUpdate()
+    {
+        InvokeAction("editBillButton");
+        return CreatePageObject<EditBillPageObject>();
+    }
```
FelixSchoberRubi, line 28: This function should be ClickEdit.
I just now got confused and thought this was the Update button inside the EditBill page.

RubeesEat.IntegrationTests/DB/DbBillRepositoryTest.cs
FelixSchoberRubi, line 1: In InMemoryBillRepositoryTest you added a test for an invalid Bill, but here it is missing

RubeesEat/Pages/EditBill.cshtml
FelixSchoberRubi, line 1: This file has a lot of similarities to SplitBill. Is there a way to avoid code duplication here? Maybe discuss with Lukas during standup

RubeesEat/wwwroot/js/editableBill.js
```diff
@@ -1,12 +1,76 @@
...
+        counter++;
+
+
+    };
+}
+
+document.addEventListener('DOMContentLoaded', () => {
+    console.log(existingEntryLines);
+    addExistingEntries(existingEntryLines);  
+});
```
FelixSchoberRubi, line 66: This has a lot of the same code as AddPerson. Maybe you can seperate the common code into AddPerson, and move the code that is specific into 2 other functions (so one function AddExistingEntrylines and one AddPersonByDropdown, which both use the AddPerson function for the code that is used in both)

RubeesEat/wwwroot/js/editableBill.js
```diff
@@ -58,7 +122,24 @@ function AddPerson() {
...
+    addToOptions.value = personName;
+    addToOptions.dataset.id = personId;
+    addToOptions.textContent = personName;
+    selectElement.appendChild(addToOptions);
+
+    const addPersonDiv = document.getElementById("addPerson");
+    addPersonDiv.hidden = false;
+}
+
+
```
FelixSchoberRubi, line 141: This function doesn't seem to be used anywhere

RubeesEat/Pages/EditBill.cshtml
```diff
@@ -0,0 +1,65 @@
...
+                                var entryLine = currentBill.EntryLines[i];
+                                if (person.Id == entryLine.Person.Id)  
+                                    checkIfPersonAlreadySelected = true;
+                            }
+
+                            if (!checkIfPersonAlreadySelected)
+                            {
+                                <option value="@person.FirstName @person.LastName" data-id="@person.Id">@person.FirstName @person.LastName </option>
+                            }
+                        }
```
FelixSchoberRubi, line 51:
```suggestion
@foreach (Person person in PersonRepository.GetAllActive().Except(currentBill.EntryLines.Select(entryLine => entryLine.Person).ToList()))
                        {
                            <option value="@person.FirstName @person.LastName" data-id="@person.Id">@person.FirstName @person.LastName </option>
                        }
```

RubeesEat/Model/DB/DbBillRepository.cs
```diff
@@ -64,6 +64,41 @@ public void Add(Bill bill)
...
+                    entryLineCommand.AddParameter($"@PersonID{i}", entryLine.Person.Id);
+                    entryLineCommand.AddParameter($"@EntryLineBillID{i}", bill.Id);
+                    entryLineCommand.ExecuteNonQuery();
+                }
+                
+                transaction.Commit();
+            }
+        }
+    }
+
```
FelixSchoberRubi, line 101: I think this is missing the exception for when a billid is not found

RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
FelixSchoberRubi, line 1: Tests for the different Bad Requests for Update in BillController are missing

RubeesEat.IntegrationTests/SeleniumTests/EditBillSeleniumTests.cs
FelixSchoberRubi, line 1: This should have the other tests that SplitBill has. Unless we find a way for the SplitBill and EditBill page to no longer have duplicated code, then we might be able to have less tests

FelixSchoberRubi (changes requested):

RubeesEat.IntegrationTests/PageObjects/EditBillPageObject.cs
```diff
@@ -0,0 +1,61 @@
...
+
+    public List<string> GetSelection()
+    {
+        var selectElement = FindElement("billPeople");
+        var options = selectElement.FindElements(By.TagName("option"));
+        return options.Select(e => e.Text).ToList();
+    }
+    
+    
+}
```
FelixSchoberRubi, line 61: This is very similar to SplitBillPageObject. I think the only difference is which button can be pushed to submit.
We should create a more general PageObject Class, of which SplitBillPageObject and EditBillPageObject derive.
Or we can simply rename both buttons into submitBill. That way we only need one PageObject class which works for both

FelixSchoberRubi (changes requested):

a-ctor (approved):

RubeesEat/Model/DB/DbBillRepository.cs
```diff
@@ -64,6 +64,44 @@ public void Add(Bill bill)
...
+            
+            GetById(bill.Id);
+            
+            using var transaction = (MySqlTransaction)connection.BeginTransaction(IsolationLevel.ReadCommitted);
+            {
+                var command = transaction.CreateCommand(
+                    $"UPDATE Bills SET Description = @Description WHERE BillID = @BillId;")!;
+                command.AddParameter("@BillID", bill.Id);
+                command.AddParameter("@Description", bill.Description);
+                command.ExecuteNonQuery();
```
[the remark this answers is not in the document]
KuJo7: We will do that in a different task (DSEAT-59)

PR #47, opened 2024-10-18
DSEAT-51 Show last people included in bill

KuJo7:
https://re-motion.atlassian.net/browse/DSEAT-51

### DIFF
--- a/RubeesEat.IntegrationTests/PageObjects/PersonButtonControlObject.cs
+++ b/RubeesEat.IntegrationTests/PageObjects/PersonButtonControlObject.cs
@@ -1,0 +1,16 @@
+using OpenQA.Selenium;
+using RubeesEat.IntegrationTests.WebTesting;
+
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public class PersonButtonControlObject : ControlObject
+{
+    private readonly PageObject _page;
+
+    public PersonButtonControlObject(PageObject page, IWebElement me) : base(page, me)
+    {
+        _page = page;
+    }
+    
+    public string Text => Me.Text;
+}
--- a/RubeesEat.IntegrationTests/PageObjects/RecentPeopleControlObject.cs
+++ b/RubeesEat.IntegrationTests/PageObjects/RecentPeopleControlObject.cs
@@ -1,0 +1,21 @@
+using OpenQA.Selenium;
+using RubeesEat.IntegrationTests.WebTesting;
+
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public class RecentPeopleControlObject : ControlObject
+{
+    private PageObject _page;
+
+    public RecentPeopleControlObject(PageObject page, IWebElement me) : base(page, me)
+    {
+        _page = page;
+    }
+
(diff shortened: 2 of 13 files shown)


FelixSchoberRubi (approved):

RubeesEat.IntegrationTests/PageObjects/PersonButtonControlObject.cs
```diff
@@ -0,0 +1,16 @@
...
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public class PersonButtonControlObject : ControlObject
+{
+    private readonly PageObject _page;
+
+    public PersonButtonControlObject(PageObject page, IWebElement me) : base(page, me)
+    {
+        _page = page;
+    }
```
a-ctor, line 13: We don't need to save the `PageObject` in a field as it is already done by the base class.

RubeesEat.IntegrationTests/PageObjects/SpiltBillPageObject.cs
```diff
@@ -45,11 +45,23 @@ public PersonAmountControlObject[] GetPersonAmounts()
...
     public List<string> GetSelection()
     {
         var selectElement = FindElement("billPeople");
         var options = selectElement.FindElements(By.TagName("option"));
         return options.Select(e => e.Text).ToList();
     }
+
+    public RecentPeopleControlObject GetRecentPeople()
+    {
+        return new RecentPeopleControlObject(this, FindElement("recentPeople"));
```
a-ctor, line 65: We shouldn't create a control object if it has the same target as the current page object. We can get rid of the control object and to the one method here inline. So this will return `PersonButtonControlObject[]` directly.

RubeesEat.IntegrationTests/PageObjects/RecentPeopleControlObject.cs
```diff
@@ -0,0 +1,21 @@
+using OpenQA.Selenium;
+using RubeesEat.IntegrationTests.WebTesting;
+
+namespace RubeesEat.IntegrationTests.PageObjects;
+
+public class RecentPeopleControlObject : ControlObject
```
a-ctor, line 6: No longer necessary if we do the refactor from the other comment.

RubeesEat/Model/IBillRepository.cs
```diff
@@ -8,4 +8,5 @@ public interface IBillRepository
     PaginatedView<BalanceChange> GetRecentBalanceChanges(Person currentUser, int page, int pageSize);
     decimal GetBalance(Person user);
     Bill? GetById(Guid guid);
+    public IReadOnlyList<Person> GetRecentBillParticipants(Person user, int limit);
```
a-ctor, line 11:
```suggestion
    IReadOnlyList<Person> GetRecentBillParticipants(Person user, int limit);
```

a-ctor (commented):
Please rebase to fix the merge conflicts in the HTML and JS. Once that is done I will review those files as well.

KuJo7:
Some errors in js are still left and needs to be fixed

PR #50, opened 2024-10-29
Feature/dseat 59 split bill edit date of bill

KuJo7:
https://re-motion.atlassian.net/browse/DSEAT-59

### DIFF
--- a/RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
+++ b/RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
@@ -18,6 +18,7 @@ public class BillControllerTest : IntegrationTestBase
         var content = new FormUrlEncodedContent(new[]
         {
             new KeyValuePair<string, string>("billDescription", "test234"),
+            new KeyValuePair<string, string>("billDate", "2024-10-29"),
             new KeyValuePair<string, string>("billAmount", "10"),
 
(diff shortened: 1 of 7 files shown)


RubeesEat.IntegrationTests/Controllers/BillControllerTests.cs
```diff
@@ -18,6 +18,7 @@ public class BillControllerTest : IntegrationTestBase
         var content = new FormUrlEncodedContent(new[]
         {
             new KeyValuePair<string, string>("billDescription", "test234"),
+            new KeyValuePair<string, string>("billDate", DateTime.Today.ToString("yyyy-MM-dd")),
```
a-ctor, line 21: We should use a fixed date here and assert that this date was actually used when creating the ticket. Especially with dates it is easy to default to the current date, which is not want we want here.

RubeesEat/Model/Bill.cs
```diff
@@ -7,6 +7,9 @@ public class Bill
     public static Bill Create(string description, EntryLine[] entryLines) =>
         new(Guid.NewGuid(), DateTime.Now, description, [..entryLines]);
```
a-ctor, line 8: Please chain this to the new method with the date parameter

a-ctor (approved):

